PR #3, opened 2024-04-29
Use cartesi/sdk for anvil/devnet at run command

endersonmaia:


endersonmaia:
Deployment with `cast` is failing.

tuler:
cast actually works for me.

My error is:
```
ERR Error: StateServerError { source: TonicError { context: "`subscribe_blocks` request", source: Status { code: Unavailable, message: "BlockUnavailable", metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Mon, 29 Apr 2024 19:46:27 GMT", "content-length": "0"} }, source: None } } } service=dispatcher
```

.changeset/healthy-windows-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cartesi/cli": patch
```
tuler, line 2: Should be minor

tuler:
The issue seems to be related to upgrading anvil.
https://github.com/foundry-rs/foundry/issues/7502
When we load a state the starting block number is 22.
When the node tries to subscribe to a block earlier than that it breaks.

The node team suggested to update the genesis block to 22.
But it's still not working.

Going back to an old anvil version, like the one used by sunodo 0.13.1, works.

endersonmaia:
> Going back to an old anvil version, like the one used by sunodo 0.13.1, works.

Opened a PR for that #4

tuler (approved):
LGTM. Squash fixup

tuler (approved):

PR #12, opened 2024-05-02
Add cruntime package

endersonmaia:
This will make it possible for Cartesi application developer to create their application Dockerfile without the need to install machine-emulator-tools and its dependencies but just the application code itself.

For ex., the following `Dockerfile` could be used for a javascript application:

```Dockerfile
FROM node:20.8.0-bookworm as build-stage
WORKDIR /opt/cartesi/dapp
COPY . .
RUN yarn install && yarn build

FROM --platform=linux/riscv64 cartesi/node:20.8.0-jammy-slim
WORKDIR /opt/cartesi/dapp
COPY --from=build-stage /opt/cartesi/dapp/dist .
CMD ["node", "index.js"]
```

How to test:

```
pnpm install && pnpm build
docker image tag cartesi/cruntime:devel cartesi/cruntime:0.1.0
alias cartesi-dev=$PWD/apps/cli/bin/run.js
```

With that, you can build an application with the sample Dockerfile above using

```
cartesi-dev build
cartesi-dev run
cartesi-dev send
```

You could also try running with `cartesi-dev --no-backend` and running the application container with docker:

```
docker build -t jsdapp . -f Dockerfile
cartesi-dev --no-backend
docker run -ti --rm -e ROLLUP_HTTP_SERVER_URL=http://localhost:8080/host-runner --network=host jsdapp
cartesi-dev send
```

---

Details:

We used `chisel` to make a small runtime image with `crun` and its dependencies.

We compiled `crun` ourselves to use the latest `crun` version and remove what we don't need.

References: 

- https://governance.cartesi.io/t/provide-an-oci-runtime-inside-the-cartesi-machine/411


packages/cruntime/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+    --release /chisel-24.04 \
+    --root /rootfs \
+    --arch=riscv64 \
+    # base rootfs dependencies
+    base-files_base \
+    base-files_release-info \
+    base-passwd_data \
+    # machine-emulator-tools dependencies
+    busybox-static_bins \
+    libgcc-s1_libs \
```
tuler, line 54: is this explicitly required, or should come by transitive dependency?

packages/cruntime/Dockerfile
```diff
@@ -0,0 +1,81 @@
...
+RUN dpkg -x /machine-emulator-tools-v${MACHINE_EMULATOR_TOOLS_VERSION}.deb /rootfs
+
+# Get chisel binary
+ARG CHISEL_VERSION
+ADD "https://github.com/canonical/chisel/releases/download/v${CHISEL_VERSION}/chisel_v${CHISEL_VERSION}_linux_riscv64.tar.gz" chisel.tar.gz
+RUN tar -xvf chisel.tar.gz -C /usr/bin/
+
+# Extract crun dependencies into the chiselled filesystem
+# FIXME: remove this when crun's dependecies slices are upstream
+ADD https://github.com/endersonmaia/chisel-releases.git#14cd0571ac7646c0e2b70e3dc41855b70805d7eb /chisel-24.04
```
tuler, line 43: let's organize a fork under cartesi org

packages/cruntime/Dockerfile
endersonmaia: The installation of mahine-emulator-tools is made outside the chisel process, so we need to do this way.

packages/cruntime/Dockerfile
```diff
@@ -0,0 +1,121 @@
...
+
+###############################################################################
+# STAGE: base-image
+#
+# This stage creates a base-image with apt repository cache and ca-certificates
+# to be used by later stages.
+#
+FROM ${IMAGE_REGISTRY}/${IMAGE_NAMESPACE}/${IMAGE_NAME}:$IMAGE_TAG AS base-image
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y --no-install-recommends ca-certificates
```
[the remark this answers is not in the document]
endersonmaia: We only use this so it's possible to download packages with chisel later.

On the reproducibility side, chisel itself doesn't support defining specific versions of packages.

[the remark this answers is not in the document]
endersonmaia: I'll add the ca-certificates version anyway.

packages/cruntime/Dockerfile
endersonmaia: https://github.com/cartesi/chisel-releases

packages/cruntime/Dockerfile
tuler: But doesn't the debian installation of the tools bring what is needed?

endersonmaia: I'm not installing, but just extracting its contents.

If I install it, it's gonna pollute the `/rootfs` that should be minimal.

PR #17, opened 2024-05-06
Feat/add cruntime for machine sdk 17 use tar stream

endersonmaia:
Attempt to use tar-stream instead of nanotar.


apps/cli/src/commands/build.ts
```diff
@@ -419,29 +362,60 @@ Update your application Dockerfile using one of the templates at https://github.
...
+            const pack = tar.pack();
+            const appGnuTarStream = fs.createReadStream(appGnutarPath);
+            const appOCIBundleStream = fs.createWriteStream(appOCIBundlePath);
+            const ociConfigJSON = JSON.stringify(createConfig(imageInfo));
+
+            // add config.json
+            pack.entry({ name: "config.json" }, ociConfigJSON, function (err) {
+                if (err) throw err;
+                console.log("ERROR config.json");
+                pack.finalize();
```
tuler, line 376: this looks out of place. because you will add more entries (for each gnuTar entry)

tuler (commented):
how do I test this?
it complains about `cartesi/cruntime:0.1.0`. How do I build it?

endersonmaia:
When you build the cruntime package , you,d need to relabel :devel as :0.1.0


docker image tag cartesi/cruntime:devel cartesi/cruntime:0.1.0

Em seg., 6 de mai. de 2024, 20:10, Danilo Tuler ***@***.***>
escreveu:

apps/cli/src/commands/build.ts
endersonmaia: I did some changes in a fixup, but the `.cartesi/app.ocibuindle.tar` is still incomplete.

I don't fully understand how these streams are supposed to work, something is missing to close/finish the tar, since I never hit the `console.log()` at lines 388 and 392.

But `.cartesi/app.ocibuindle.tar` has `config.json`  and `rootfs/` prefix there.

endersonmaia:
> how do I test this? it complains about `cartesi/cruntime:0.1.0`. How do I build it?

I recommend testing the draft branch, to see the `cruntime` working, and this PR is an improvemente so we don't create a --flash-driver just for the `config.json`.

PR #24, opened 2024-05-11
feat: support for multiple sdks and abstract machine making into create_image script in sdk

stskeeps:
This probably requires upping CARTESI_DEFAULT_SDK_VERSION as well.

### DIFF
--- a/apps/cli/src/commands/build.ts
+++ b/apps/cli/src/commands/build.ts
@@ -19,6 +19,7 @@ type ImageInfo = {
     env: string[];
     ramSize: string;
     sdkVersion: string;
+    sdkName: string;
     workdir: string;
 };
 
@@ -28,7 +29,8 @@ const CARTESI_LABEL_DATA_SIZE = `${CARTESI_LABEL_PREFIX}.data_size`;
 const CARTESI_DEFAULT_RAM_SIZE = "128Mi";
 
 const CARTESI_LABEL_SDK_VERSION = `${CARTESI_LABEL_PREFIX}.sdk_version`;
-const CARTESI_DEFAULT_SDK_VERSION = "0.6.0";
+const CARTESI_LABEL_SDK_NAME = `${CARTESI_LABEL_PREFIX}.sdk_name`;
+const CARTESI_DEFAULT_SDK_VERSION = "0.7.0";
 
 export default class BuildApplication extends BaseCommand<
     typeof BuildApplication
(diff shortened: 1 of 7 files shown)


apps/cli/src/commands/build.ts
```diff
@@ -19,6 +19,7 @@ type ImageInfo = {
     env: string[];
     ramSize: string;
     sdkVersion: string;
+    sdkType: string;
```
endersonmaia, line 22: I suggest making this variable consistent with the label.

-  `sdkName`
- `CARTESI_LABEL_SDK_NAME`
- `io.cartesi.rollups.sdk_name`

packages/sdk/Dockerfile
```diff
@@ -105,6 +105,7 @@ COPY devnet /usr/local/bin
 COPY eth_isready /usr/local/bin
 COPY eth_dump /usr/local/bin
 COPY eth_load /usr/local/bin
+COPY create_image /usr/local/bin
```
endersonmaia, line 108: `create_image` can confuse with the ext2 image and others, at least for me.

Maybe `create_snapshot` or even `create_machine_snapshot`.

packages/sdk/create_image
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
```
endersonmaia, line 1:
```suggestion
#!/usr/bin/env bash

set -e
```

packages/sdk/create_image
```diff
@@ -0,0 +1,81 @@
...
+while true ; do
+    case "$1" in
+        -r|--ram-length)
+            ram_length="$2"
+            shift 2 ;;
+        -d|--drive-label)
+            drive_label="$2"
+            shift 2 ;;
+        -m|--machine-rootfs)
+            input_filename="$2"
```
endersonmaia, line 40: I'd match variable and arguments here.

```suggestion
        -m|--drive-filename)
            drive_filename="$2"
```

endersonmaia (changes requested):
Install [changesets](https://github.com/changesets/changesets) and add a changset file where you propose the minor version bump for the cartesi/sdk package.

endersonmaia (changes requested):
@stskeeps changeset will take care of updating the CHANGELOG and package.json files when changes hit the main branch, a GH Actions job will create/update the Release PR.

Since this PR has changes for two packages with a dependency between them, we'd need to have the changes in separate PRs.

The first one is with changes to the cartesi/sdk package, so we can release it first. Another PR with changes to the cartesi/cli making use of the new cartesi/sdk.

stskeeps:
OK, so close this pull request and I've opened as a start https://github.com/cartesi/cli/pull/26 - this is the style desired right?

PR #26, opened 2024-05-13
feat: add create_machine_snapshot

stskeeps:

PR #27, opened 2024-05-13
feat: allow to select sdk with .sdk_name

stskeeps:


apps/cli/src/commands/build.ts
```diff
@@ -28,7 +29,8 @@ const CARTESI_LABEL_DATA_SIZE = `${CARTESI_LABEL_PREFIX}.data_size`;
 const CARTESI_DEFAULT_RAM_SIZE = "128Mi";
 
 const CARTESI_LABEL_SDK_VERSION = `${CARTESI_LABEL_PREFIX}.sdk_version`;
-const CARTESI_DEFAULT_SDK_VERSION = "0.6.0";
+const CARTESI_LABEL_SDK_NAME = `${CARTESI_LABEL_PREFIX}.sdk_name`;
+const CARTESI_DEFAULT_SDK_VERSION = "0.7.0";
```
endersonmaia, line 33: Since the PR #26  suggests a patch to cartesi/sdk, this should be:

```suggestion
const CARTESI_DEFAULT_SDK_VERSION = "0.6.2";
```

endersonmaia (changes requested):

stskeeps:
This should now in shape, opinions?

tuler (approved):

tuler:
I tested with rollups and it works as it should. Applications will need to upgrade the SDK version 0.6.0 to 0.6.2, as we are setting the minimum required version to 0.6.2.

endersonmaia (approved):

PR #32, opened 2024-05-28
feat(cli): use cartesi as compose project name

endersonmaia:
This PT will make the docker compose use `cartesi` as project name.

This way, instead of having a random string as the prefix for containers we'll have `cartesi` as prefix.

This will make it easier for us to grab logs or execute commands on a specific container, like:

```shell
docker logs -f cartesi-validator-1
docker exec -ti cartesi-validator-1 ls -lha /
```

### DIFF
--- a/.changeset/wild-gorillas-decide.md
+++ b/.changeset/wild-gorillas-decide.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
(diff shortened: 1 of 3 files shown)


tuler:
The drawback is that if you try to run two projects simultaneously it will fail.
Running two projects may already fail because of ports, but I think we provide a way to change ports.

PR #35, opened 2024-07-04
Bump node to 1.5.0

tuler:
Also change semantics of epoch duration parameter of run to be in blocks


apps/cli/src/commands/run.ts
```diff
@@ -17,9 +17,9 @@ export default class Run extends BaseCommand<typeof Run> {
             description: "interval between blocks (in seconds)",
             default: 5,
         }),
-        "epoch-duration": Flags.integer({
-            description: "duration of an epoch (in seconds)",
-            default: 3600,
+        "epoch-length": Flags.integer({
+            description: "length of an epoch (in blocks)",
+            default: 720,
```
endersonmaia, line 22: Shouldn't the blockTime define this value?

If the default blockTime is 5 seconds, and we want to keep the 1 hour epochLenght, we should define this as 300.

```suggestion
            default: 300,
```

endersonmaia (changes requested):

apps/cli/src/commands/run.ts
tuler: Those are independent configurations. Forget about time. What matters are blocks.
If the developer wants to change any of those the options are there.

endersonmaia (approved):
LGTM

endersonmaia (approved):

PR #59, opened 2024-07-31
Account Abstraction in CLI

tuler:
This PR adds some relevant smart contract addresses to the `address-book`.
It also adds a new docker compose file that adds two additional services, the `/bundler` and the `/paymaster`.


tuler:
This creates two additional containers. I think it will be a good idea to have flags in the `cartesi run` command to enable or disable optional services.
The services we have now are:
- anvil
- postgres
- node
- bundler
- paymaster
- explorer

Some of these are required. But maybe explorer, bundler and paymaster could be optional.

apps/cli/src/node/docker-compose-aa.yaml
```diff
@@ -0,0 +1,86 @@
...
+                "--port",
+                "4337",
+                "--public-client-log-level",
+                "error",
+                "--wallet-client-log-level",
+                "error",
+                "--polling-interval",
+                "100",
+                "--enable-debug-endpoints",
+            ]
```
endersonmaia, line 34: If `alto` support environment variables, mayb it's cleaner to use them instead of CLI arguments.

Or keeping key=value in the same line:

```suggestion
            [
                "alto",
                "--entrypoints=0x5ff137d4b0fdcd49dca30c7cf57e578a026d2789,0x0000000071727De22E5E9d8BAf0edAc6f37da032",
                "--log-level=info",
                "--entrypoint-simulation-contract=0x74Cb5e4eE81b86e70f9045036a1C5477de69eE87",
                "--rpc-url=http://anvil:8545",
                "--min-executor-balance=0",
                "--utility-private-key=0xdbda1821b80551c9d65939329250298aa3472ba22feea921c0cf5d620ea67b97",
                "--executor-private-keys=0x2a871d0798f97d79848a013d4936a73bf4cc922c825d33c1cf7073dff6d409c6,0x4bbbf85ce3377467afe5d46f804f221813b2bb87f24d81f60f1fcdbf7cbf4356,0x92db14e403b83dfe3df233f83dfa3a0d7096f21ca9b0d6d6b8d88b2b4ec1564e,0x8b3a350cf5c34c9194ca85829a2df0ec3153be0318b5e2d3348e872092edffba,0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a",
                "--max-block-range=10000",
                "--safe-mode=false",
                "--port=4337",
                "--public-client-log-level=error",
                "--wallet-client-log-level=error",
                "--polling-interval=100",
                "--enable-debug-endpoints",
            ]
```

Or use YAML list instead of `[ ]`.

tuler: env var is not supported in alto

tuler:
Added disable flags to `run`

```shell
      --disable-bundler       disable bundler service
      --disable-explorer      disable explorer service
      --disable-paymaster     disable paymaster service
```

endersonmaia (approved):

PR #63, opened 2024-08-12
feat(cli): expose rollups-node prometheus metrics

endersonmaia:
This PR will export rollups-node (validator) metrics at `http://localhost:8080/metrics`.

```shell
> cartesi run
Attaching to prompt-1, validator-1
(...)
prompt-1     | Prometheus metrics at http://localhost:8080/metrics
prompt-1     | Press Ctrl+C to stop the node
```

### DIFF
--- a/.changeset/silent-bags-accept.md
+++ b/.changeset/silent-bags-accept.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+expose rollups-node prometheus metrics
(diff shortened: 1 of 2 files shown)


tuler:
Does node V2 have prometheus metrics?

endersonmaia:
> Does node V2 have prometheus metrics?

Not that know of.

endersonmaia:
rollups-node v2 doesn't have metrics yet, we'll come back with that when it's the time

PR #65, opened 2024-08-21
Add cpu and mem limits options

endersonmaia:
This PR will add `--cpus` and `--memory` options for `cartesi run` command. This way it will be possible to define these in alignment with the production deployment values and evaluate locally if those values are enough for the application to run smoothly.

Also, it was added a hidden `--dry-run` to the run command, that will show the .yaml compose file in case it's needed to debug the compose file.

### DIFF
--- a/.changeset/chatty-toes-relate.md
+++ b/.changeset/chatty-toes-relate.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+adds --dry-run to run command
--- a/apps/cli/src/commands/run.ts
+++ b/apps/cli/src/commands/run.ts
@@ -54,6 +54,16 @@ export default class Run extends BaseCommand<typeof Run> {
             description: "port to listen for incoming connections",
             default: 8080,
         }),
+        cpus: Flags.integer({
+            description:
+                "Define the number of CPUs to use (eg.: 1) for the rollups-node",
+            summary: "number of cpu limits for the rollups-node",
+        }),
+        memory: Flags.integer({
+            description:
+                "Define the amount of memory to use for the rollups-node in MB (eg.: 1024)",
+            summary: "memory limit for the rollups-node in MB",
+        }),
         "dry-run": Flags.boolean({
             description: "show the docker compose configuration",
             default: false,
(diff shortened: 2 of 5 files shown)


gbarros:
This is a 👍 change!

apps/cli/src/commands/run.ts
```diff
@@ -54,6 +54,25 @@ export default class Run extends BaseCommand<typeof Run> {
             description: "port to listen for incoming connections",
             default: 8080,
         }),
+        cpus: Flags.integer({
+            description:
+                "Define the number of CPUs to use (eg.: 1) for the rollups-node, 0 (zero) means unlimited",
+            summary: "number of cpu limits for the rollups-node",
+            default: 4,
```
tuler, line 61: I think the option should be optional, without any default value.
It will default to whatever docker compose defines as a non-defined resources: limitation.

endersonmaia: fixup sent 9c565b3

tuler: No, the default is not `0`. It should be `undefined`.

endersonmaia: The way I exposed to the cartesi cli, if you define `0` it translates to undefined, since the `resources:` section will not be loaded into the final .yaml.

You can test with `cartesi run --show-config` vs. `cartesi run --show-config --cpus=4 --memory=4096`

tuler: That is because `if(0)` is false in typescript, but it's more correct to explicitly be undefined.
`0` is different from `undefined`.

endersonmaia: sent fixup 4c493a1

tuler:
I see that the `show-config` option only shows the config, and don't actually run the service.
What do you think of calling the option `--dry-run`?

tuler:
```shell
cartesi run --memory 512 --show-config
```

The command above produced the following output:

```yaml
    deploy:
      resources:
        limits:
          memory: "536870912"
```

I don't know if that value is correct, as it should be a "byte value".
https://docs.docker.com/reference/compose-file/extension/#specifying-byte-values

endersonmaia:
> ```shell
> cartesi run --memory 512 --show-config
> ```
> 
> The command above produced the following output:
> 
> ```yaml
>     deploy:
>       resources:
>         limits:
>           memory: "536870912"
> ```
> 
> I don't know if that value is correct, as it should be a "byte value". https://docs.docker.com/reference/compose-file/extension/#specifying-byte-values

The output is after the processing of the values. 

The `cartesi run --memory` option should be in MB.

.changeset/chatty-toes-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+adds --show-config to run command
```
tuler, line 5:
```suggestion
adds --dry-run to run command
```

tuler (approved):

PR #73, opened 2024-09-09
cli: enforce build for linux/riscv64 platform

endersonmaia:
Since the target is always a Cartesi Machine (riscv64), cartesi build should enforce the docker build environment via `--platform` argument.

This way `Dockerfile` from cartesi/application-templates should be something like this.

**javascript**

```Dockerfile
#BUILDER
FROM --platform=$BUILDPLATFORM node:20.16.0-bookworm AS build-stage
(...)

#TARGET
FROM cartesi/node:20.16.0-jammy-slim
(...)
```

We should only need to define `--platform` when we want to run something on the host platform, for a cross-compilation or build stage that could be faster without the linux/riscv64 qemu of the docker builder.

The example above uses the `$BUILDPLATFORM`, which is recommended so it will use the host platform.

The second and last `FROM` doesn't need a `--platform` since this will be enforced by the `cartesi build`.

### DIFF
--- a/.changeset/big-ghosts-rescue.md
+++ b/.changeset/big-ghosts-rescue.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+enforce build for linux/riscv64 platform
--- a/apps/cli/src/builder/docker.ts
+++ b/apps/cli/src/builder/docker.ts
@@ -26,6 +26,8 @@ const buildImage = async (options: ImageBuildOptions): Promise<string> => {
     const args = [
         "buildx",
         "build",
+        "--platform",
+        "linux/riscv64",
         "--file",
         dockerfile,
         "--load",
--- a/apps/cli/src/builder/docker.ts
+++ b/apps/cli/src/builder/docker.ts
@@ -53,6 +53,9 @@ const buildImage = async (options: ImageBuildOptions): Promise<string> => {
  * @returns Information about the image
  */
 const getImageInfo = async (image: string): Promise<ImageInfo> => {
+    // pull image to ensure it is available
+    await execa("docker", ["image", "pull", image]);
+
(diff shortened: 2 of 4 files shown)


.changeset/big-ghosts-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+enforce build for linux/riscv64 paltform
```
tuler, line 5: typo

tuler:
I think this will break compatibility with current Dockerfiles out there, right?

endersonmaia:
> I think this will break compatibility with current Dockerfiles out there, right?

Yes. :thinking: 

What do you suggest? make it a `minor` ? 

And we should also update the applications-templates.

.changeset/big-ghosts-rescue.md
endersonmaia: even the branch name has a type :facepalm: 

I've sent a fix with push force to fix the type in the changeset and commit comment

tuler:
> What do you suggest? make it a `minor` ?

I think it would be a minor.

tuler:
As this breaks Dockerfile compatibility let's put it together with other changes that changes Dockerfile, like the v2 initiative.

.changeset/big-ghosts-rescue.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cartesi/cli": patch
```
tuler, line 2: in pre-release mode should we specify as major or as patch?
because it's meant to be `2.0.0-alpha.1`

endersonmaia: I'm considering this particular commit as a patch. I think everything that goes into prerelease will bump the last number after `alpha` only.

tuler (approved):

PR #74, opened 2024-09-18
Bump e2fsprogs

tuler:
This PR install a e2fsprogs from bookworm-backports, which provides support for building drives from tar files

### DIFF
--- a/.changeset/ninety-countries-complain.md
+++ b/.changeset/ninety-countries-complain.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/sdk": minor
+---
+
+bump e2fsprogs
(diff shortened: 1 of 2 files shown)


tuler:
This will actually need to go to main so we can release an SDK with it

endersonmaia:
> This will actually need to go to main so we can release an SDK with it

So update the target branch for the PR, if possible.

I see this commit already exists on the other PR I just reviewed and approved #75

tuler:
Changed to minor version
Removed some lint warnings

endersonmaia (changes requested):
Use an individual commit for the linting fixes, and squash the first 2 commits into one.

endersonmaia (approved):

PR #75, opened 2024-09-18
Lint and cosmetic changes

tuler:
- delete unused tests
- delete unused var
- delete unused type


endersonmaia (approved):
While at it, you could also fix all Dockerfile warning we're getting at GitHub.

It's up to you do this on this PR or another.

PR #78, opened 2024-09-18
Yaml configuration

tuler:


apps/cli/src/commands/build.ts
```diff
@@ -1,312 +1,645 @@
...
+ * @returns return of execa
+ */
+type OptionsDockerFallback = Options & { image?: string };
+const execaDockerFallback = async (
+    command: string,
+    args: readonly string[],
+    options: OptionsDockerFallback,
+) => {
+    try {
+        return execa(command, args, options);
```
endersonmaia, line 51:
```suggestion
        const result = await execa(command, args, options);
        return result;
```

apps/cli/src/commands/build.ts
```diff
@@ -1,312 +1,645 @@
...
+    } catch (error) {
+        if (error instanceof ExecaError) {
+            if (error.code === "ENOENT" && options.image) {
+                return execa(
```
endersonmaia, line 55:
```suggestion
                return await execa(
```

apps/cli/src/commands/build.ts
```diff
@@ -1,312 +1,645 @@
...
+    const PATH = "/opt/homebrew/opt/e2fsprogs/sbin";
+    const args = [
+        "-b",
+        blockSize.toString(),
+        "-U",
+        "clear",
+        "-E",
+        `hash_seed=${seed}`,
+        "-d",
+        input,
```
endersonmaia, line 408: Neither my local `mke2fs` (Ubuntu 24.04 `1.47.0 (5-Feb-2023)`) nor the one installed on cartesi/sdk accepts a tarball for `-d` argument.

So we'd need to unpack the tarball into a directory before at this `tarToExt()` function.

Or we should compile mke2fs inside cartesi/sdk with `-d` support to tarball, so that `execaDockerFallback()` doesn't fail right ahead.

endersonmaia (changes requested):
There's a lot of changes, are you planning to split those in smaller commits or PRs?

I was just exploring the build process with `mke2fs` but there's a lot more to review and test, and this is going to be hard with such a big PR.

apps/cli/src/commands/build.ts
endersonmaia: Only available on e2fsprogs >= 1.47.1

>   * Mke2fs -d can now support an input tar file if the libarchive library
    is available.

tuler: That was why I changed the sdk to include version 1.47.1

endersonmaia: Yeah, and it was failing because this PR ignores the `LABEL`  inside `Docekrfile` and wasn't using the `cartesi/sdk:devel` but `0.10.0` version

it worked with `devel`, so I'm resolving this conversation

tuler:
Maybe we can merge this, and retry to organize the PR?

PR #83, opened 2024-10-10
Shell with config based machine

tuler:
This starts the configured machine with a shell

### DIFF
--- a/apps/cli/src/commands/shell.ts
+++ b/apps/cli/src/commands/shell.ts
@@ -1,9 +1,9 @@
 import { Args, Flags } from "@oclif/core";
-import { execa } from "execa";
 import fs from "fs-extra";
-import { lookpath } from "lookpath";
 import path from "path";
 import { BaseCommand } from "../baseCommand.js";
+import { ImageInfo } from "../config.js";
+import { bootMachine } from "../machine.js";
 
 export default class Shell extends BaseCommand<typeof Shell> {
     static description = "Start a shell in cartesi machine of application";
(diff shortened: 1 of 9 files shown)


apps/cli/src/commands/shell.ts
```diff
@@ -18,62 +18,64 @@ export default class Shell extends BaseCommand<typeof Shell> {
...
+            cmd: [],
+            entrypoint: ["/bin/bash"],
+            env: [],
+            workdir: "/",
+        };
+
+        // start with interactive mode on
+        config.machine.interactive = true;
+
+        /* why this?
```
endersonmaia, line 61: `-it` is the enhanced mode and requires the `stty` binary

```
 -i or --htif-console-getchar
    run in interactive mode using a HTIF console device.

    NON REPRODUCIBLE OPTION, DON'T USE THIS OPTION IN PRODUCTION
(...)
  -it
    run in enhanced interactive mode using a VirtIO console device.
    the console is resizable, more responsive, and support more features
    than the -i option.

    like --virtio-console, but automatically appends init commands to forward
    TERM and LANG environment variables from the host to the guest,
    allowing the use of true colors and special characters (when supported).
```

apps/cli/src/machine.ts
```diff
@@ -0,0 +1,103 @@
...
+        args.push("--assert-rolling-template");
+    }
+    if (finalHash) {
+        args.push("--final-hash");
+    }
+    if (info?.workdir) {
+        args.push(`--append-init=WORKDIR="${info.workdir}"`);
+    }
+    if (interactive) {
+        args.push("-it");
```
endersonmaia, line 83: we should check for `stty` like we did before

tuler: Is `stty` not installed in any platform?

endersonmaia: > What happens if this stty is not installed?

I don't know. And I'm not finding the reference I used when I implemented this.

tuler: Did you need to install `stty`?

endersonmaia: > Did you need to install stty?

No

apps/cli/src/commands/shell.ts
```diff
@@ -18,62 +18,66 @@ export default class Shell extends BaseCommand<typeof Shell> {
...
+                throw new Error(
+                    `drive '${name}' not built, run '${this.config.bin} build'`,
+                );
+            }
         }
 
+        // create shell entrypoint
+        const info: ImageInfo = {
+            cmd: [],
+            entrypoint: ["/bin/bash"],
```
endersonmaia, line 55:
```suggestion
            entrypoint: ["/bin/sh"],
```

endersonmaia (changes requested):

apps/cli/src/commands/shell.ts
```diff
@@ -24,8 +24,14 @@ export default class Shell extends BaseCommand<typeof Shell> {
...
         "run-as-root": Flags.boolean({
-            description: "run as root user",
             default: false,
+            description: "run as root user",
+            summary: "run the cartesi machine as the root user",
+        }),
+        shell: Flags.string({
+            default: "/bin/sh",
+            description: "shell command to run",
+            summary: "shell to run",
```
endersonmaia, line 34:
```suggestion
        command: Flags.string({
            default: "/bin/sh",
            description: "command to run",
            summary: "command to run",
            char: "C",
```

Won't it be strang `cartesi shell --shell=/bin/bash`  ?

I think that a `--command` would be better, and it's like `flyctl` does with `flyctl ssh console -C | --command`.

endersonmaia (changes requested):

endersonmaia (approved):

PR #85, opened 2024-10-14
Add typed errors do config module

endersonmaia:
This shall make it easier to implement new tests to the config module.

### DIFF
--- a/apps/cli/src/config.ts
+++ b/apps/cli/src/config.ts
@@ -7,56 +7,56 @@ import { TomlPrimitive, parse as parseToml } from "smol-toml";
  * Typed Errors
  */
 export class InvalidBuilderError extends Error {
-    constructor(builder: string) {
+    constructor(builder: TomlPrimitive) {
         super(`Invalid builder: ${builder}`);
         this.name = "InvalidBuilder";
     }
 }
 
 export class InvalidDriveFormatError extends Error {
-    constructor(format: string) {
+    constructor(format: TomlPrimitive) {
         super(`Invalid drive format: ${format}`);
         this.name = "InvalidDriveFormatError";
     }
 }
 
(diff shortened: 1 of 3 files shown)


apps/cli/src/config.ts
```diff
@@ -3,6 +3,72 @@ import os from "os";
 import { extname } from "path";
 import { TomlPrimitive, parse as parseToml } from "smol-toml";
 
+/**
+ * Typed Errors
+ */
+export class InvalidBuilderError extends Error {
+    constructor(builder: any) {
```
tuler, line 10: don't use any, use `string`

apps/cli/src/config.ts
```diff
@@ -3,6 +3,72 @@ import os from "os";
...
+        super(`Invalid builder: ${builder}`);
+        this.name = "InvalidBuilder";
+    }
+}
+
+export class InvalidDriveFormatError extends Error {
+    constructor(format: any) {
```
tuler, line 17: don't use any, use `string`

apps/cli/src/config.ts
```diff
@@ -3,6 +3,72 @@ import os from "os";
...
+        super(`Invalid drive format: ${format}`);
+        this.name = "InvalidDriveFormatError";
+    }
+}
+
+export class InvalidEmptyDriveFormatError extends Error {
+    constructor(format: any) {
```
tuler, line 24: don't use any, use `string`

tuler (commented):
review the usage of `any` in general

apps/cli/src/config.ts
endersonmaia: When using `string`, the IDE suggest to call it likle `new InvalidBuilderError(true as unknown as string)`, which is weird for me.

What's the best approach in those cases?

tuler (approved):

PR #91, opened 2024-10-17
Bump dependencies and 1.0 release

tuler:
Bump dependencies
1.0 release


endersonmaia:
I think we need a dummy test so the CI can pass.

endersonmaia (approved):

PR #106, opened 2024-10-18
Refactor release workflow

tuler:
`cli`, `devnet`, and `mock-verifying-paymaster` are npm packages published to the `@cartesi` npm organization.
Instead of publishing the package in a workflow_call from the `release` workflow, this changes the publishing to be done in the `release` workflow itself in the `Create Release Pull Request or Publish to npm` step.

The publishing of the `sdk` docker image is still done in a workflow_call.

This will fix the problem that happens in the pre-release scenario, because `changeset publish` handles the case of specifying the npm tag to `pnpm publish` automatically.

### DIFF
--- a/.github/workflows/cli.yaml
+++ b/.github/workflows/cli.yaml
@@ -1,15 +1,5 @@
 name: cli
 on:
-    workflow_call:
-        inputs:
-            release:
-                type: boolean
-                required: false
-        secrets:
-            NPM_TOKEN:
-                required: true
-            TURBO_TOKEN:
-                required: true
     pull_request:
         paths:
             - .github/workflows/cli.yaml
@@ -48,22 +38,14 @@ jobs:
             - name: Install dependencies
               run: pnpm install
 
-            - name: Test / Coverage
-              working-directory: ./apps/cli
-              run: pnpm run test:coverage
+            - name: Build
+              run: pnpm build --filter @cartesi/cli
+
+            - name: Test
+              run: pnpm test --filter @cartesi/cli
 
             - name: "Report Coverage"
               if: always()
               uses: davelosert/vitest-coverage-report-action@v2
               with:
                   working-directory: ./apps/cli
-
-            - name: Build
-              run: pnpm build --filter @cartesi/cli
-
-            - name: Publish
-              if: ${{ inputs.release }}
-              working-directory: ./apps/cli
-              run: pnpm publish --access public
-              env:
-                  NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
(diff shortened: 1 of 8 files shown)


endersonmaia (approved):

.github/workflows/cli.yaml
```diff
@@ -48,22 +38,14 @@ jobs:
...
               run: pnpm install
 
-            - name: Test / Coverage
-              working-directory: ./apps/cli
-              run: pnpm run test:coverage
+            - name: Build
+              run: pnpm build --filter @cartesi/cli
+
+            - name: Test
+              run: pnpm test --filter @cartesi/cli
```
endersonmaia, line 45: The next step will require that the test:covarega is executed.

endersonmaia: Maybe it's better to run inside apps/cli to ensure the coverage is in the right place instead of using --filter ?

.github/workflows/cli.yaml
```diff
@@ -48,22 +38,14 @@ jobs:
...
 
             - name: "Report Coverage"
               if: always()
               uses: davelosert/vitest-coverage-report-action@v2
               with:
                   working-directory: ./apps/cli
```
endersonmaia, line 51: Like I commented before, does running with filter will create the coverage/ directory at the root or inside apps/cli ?

.github/workflows/release.yaml
```diff
@@ -38,14 +38,27 @@ jobs:
             - name: Install Dependencies
               run: pnpm install
 
-            - name: Create Release Pull Request
+            - name: Build
+              run: pnpm build
+
+            - name: Test
+              run: pnpm test
```
endersonmaia, line 45: This should be test:coverage só that the next step has the data.

Have to check if there's a vitest config at this level that generates the json coverage report

apps/cli/vitest.config.ts
```diff
@@ -3,6 +3,7 @@ import { defineConfig } from "vitest/config";
 export default defineConfig({
     test: {
         coverage: {
+            enabled: true,
```
endersonmaia, line 6: Ok, ignore previous config:coverage comments

endersonmaia: But still, this config will only aplly to apps/cli AFAIK.

endersonmaia (changes requested):

.github/workflows/cli.yaml
tuler: I enabled the coverage in the vitest config file, so it's always on now

.github/workflows/cli.yaml
tuler: inside the cli. turbo basically runs tests recursively in all projects

apps/cli/vitest.config.ts
tuler: only cli has coverage test anyway

endersonmaia (approved):

endersonmaia (approved):

PR #107, opened 2024-10-21
Tools exec wrappers

tuler:
This creates wrappers for the external tools we use in the CLI.
Each publishes a `version` method to report a `SemVer`, which we can use to check the minimum version in the execution itself, or in a `doctor command`.
I've setup the integration tests to use the `cartesi/sdk:0.11.0` docker image as fallback, as I'm not sure what is installed in the CI environment.

### DIFF
--- a/apps/cli/src/exec/cartesi-machine.ts
+++ b/apps/cli/src/exec/cartesi-machine.ts
@@ -1,0 +1,35 @@
+import { parse, Range, SemVer } from "semver";
+import {
+    DockerFallbackOptions,
+    execaDockerFallback,
+    ExecaOptionsDockerFallback,
+} from "./util.js";
+
+export const requiredVersion = new Range("^0.18.1");
+
+export const boot = async (
+    args: readonly string[],
+    options: ExecaOptionsDockerFallback,
+) => {
+    return execaDockerFallback("cartesi-machine", args, options);
+};
+
+export const version = async (
+    options?: DockerFallbackOptions,
+): Promise<SemVer | null> => {
+    const { image } = options || {};
+    try {
+        const { stdout } = await execaDockerFallback(
+            "cartesi-machine",
+            ["--version-json"],
+            { image },
+        );
+        if (typeof stdout === "string") {
+            const output = JSON.parse(stdout);
+            return parse(output.version);
+        }
+        return null;
+    } catch (e: unknown) {
+        return null;
+    }
+};
--- a/apps/cli/src/exec/genext2fs.ts
+++ b/apps/cli/src/exec/genext2fs.ts
@@ -1,0 +1,67 @@
+import { parse, Range, SemVer } from "semver";
+import { DockerFallbackOptions, execaDockerFallback } from "./util.js";
+
+const BLOCK_SIZE = 4096; // fixed at 4k
+
+export const requiredVersion: Range = new Range("^1.5.6");
+
+const baseArgs = (options: { extraBlocks: number }) => [
+    "--block-size",
+    BLOCK_SIZE.toString(),
+    "--faketime",
+    "--readjustment",
+    `+${options.extraBlocks}`,
+];
+
+export const fromDirectory = async (
+    options: {
+        cwd?: string;
+        extraSize: number;
+        input: string;
+        output: string;
+    } & DockerFallbackOptions,
+) => {
+    const { cwd, extraSize, image, input, output } = options;
+    const extraBlocks = Math.ceil(extraSize / BLOCK_SIZE);
+    return execaDockerFallback(
+        "xgenext2fs",
+        [...baseArgs({ extraBlocks }), "--root", input, output],
+        { cwd, image },
+    );
+};
+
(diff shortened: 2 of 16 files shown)


apps/cli/src/exec/genext2fs.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    const args = [
+        "--block-size",
+        blockSize.toString(),
+        "--faketime",
+        "--tarball",
+        input,
+        "--readjustment",
+        `+${extraBlocks}`,
+        output,
```
endersonmaia, line 46: This is very similar to the arguments from `fromDirectory`, maybe we could reduce this duplication and jave a `baseArgs` somewhere  that could be just appended with what is particular to `fromTar` and `fromDirectory`.

apps/cli/src/exec/mksquashfs.ts
```diff
@@ -0,0 +1,77 @@
...
+        "-",
+        output,
+        "-tar",
+        "-all-time",
+        "0",
+        "-all-root", // XXX: should we use this?
+        "-noappend",
+        "-comp",
+        compression,
+        "-no-progress",
```
endersonmaia, line 46: Same as previous comment for `genext2fs` module.

endersonmaia (changes requested):
To improve test coverage, we should be able to force the docker execution even if we have the binary on the host system.

```typescript
export type ExecaOptionsDockerFallback = Options & { image?: string, dockerOnly?: bool };
```

apps/cli/src/exec/cartesi-machine.ts
```diff
@@ -0,0 +1,29 @@
...
+export const boot = async (
+    args: readonly string[],
+    options: ExecaOptionsDockerFallback,
+) => {
+    return execaDockerFallback("cartesi-machine", args, options);
+};
+
+export const version = async (options?: {
+    image?: string;
+}): Promise<SemVer | null> => {
```
endersonmaia, line 13: Can we have a `REQUIRED_VERSION` for each `exec/<module>` and if `<module>.version` doesn't match the required onde we throw a typed exception.

This way we could test the semver using minor or patch when possible, without needing an exact match in case it works for the `<module>` being used.

endersonmaia (changes requested):

apps/cli/src/exec/mksquashfs.ts
tuler: done

apps/cli/src/exec/genext2fs.ts
tuler: done

tuler:
> we should be able to force the docker execution

done

apps/cli/src/exec/cartesi-machine.ts
tuler: you mean before the execution of the command?

endersonmaia: Not necessarily,  but just that the version required definition is part of the
<module> code and not the test code.

tuler: I didn't understand your request

endersonmaia: something in those lines:

```patch
diff --git a/apps/cli/src/exec/crane.ts b/apps/cli/src/exec/crane.ts
index 15a402e..13c1bcf 100644
--- a/apps/cli/src/exec/crane.ts
+++ b/apps/cli/src/exec/crane.ts
@@ -2,6 +2,8 @@ import { parse, SemVer } from "semver";
 import { Stream } from "stream";
 import { spawnSyncDockerFallback } from "./util";

+export const RequiredVersion: SemVer = parse("0.19.1")!;
+
 export const exportImage = async (options: {
     stdin: Stream;
     stdout: Stream;
diff --git a/apps/cli/test/exec/crane.test.ts b/apps/cli/test/exec/crane.test.ts
index 50021ff..3638854 100644
--- a/apps/cli/test/exec/crane.test.ts
+++ b/apps/cli/test/exec/crane.test.ts
@@ -4,6 +4,6 @@ import { crane } from "../../src/exec/index.js";
 describe("crane", () => {
     it("should report version", async () => {
         const version = await crane.version({ image: "cartesi/sdk:0.11.0" });
-        expect(version?.format()).toEqual("0.19.1");
+        expect(version?.format()).toEqual(crane.RequiredVersion.format());
     });
 });
```

endersonmaia (approved):
Approving so we move one, but we stiil need to use the exec wrapper elsewhere in the code, right?

apps/cli/src/exec/cartesi-machine.ts
tuler: done

tuler:
> Approving so we move one, but we stiil need to use the exec wrapper elsewhere in the code, right?

Right.
I also have an idea to maybe drop the integration tests of driving building, and do as integration tests of genext2fs, mksquashfs, etc, directly

PR #108, opened 2024-10-22
fix: workaround assert rolling template

stskeeps:


endersonmaia:
@tuler sholud we wait it to be fixed upstream (https://github.com/cartesi/machine-emulator/pull/284) or merge this hotfix?

this shall require patch release for sdk and cli

tuler:
I understand this only affects emulator 0.18.1, not 0.16.1, right?

endersonmaia:
> > I understand this only affects emulator 0.18.1, not 0.16.1, right?
> 
> correct, this was broken with new CMIO so

so this should target the next cli release at `prerelease/v2-alpha` branch

tuler:
> so this should target the next cli release at `prerelease/v2-alpha` branch

No, we already released SDK with emulator 0.18.1.
This fix can go to main under a SDK patch.

PR #116, opened 2024-10-29
Cartesi Rollups v2

endersonmaia:
This is support for Rollups V2.
The `run` command has been deprecated and replaced by two separate steps: `rollups start` and `rollups deploy`.

The following are the new commands:

```
$ cartesi rollups --help
Usage: cartesi rollups [options] [command]

Options:
  -h, --help        display help for command

Commands:
  start [options]   Start a local rollups node environment.
  status [options]  Shows the status of a local rollups node environment.
  stop [options]    Stop a local rollups node environment.
  deploy [options]  Deploy a rollups application to a rollups node.
```

ps: `deploy` (not `rollups deploy`) will probably be deprecated. Or we may use a `cartesi.toml` property to redirect commands from `<root>` to either `rollups` or (upcoming) `coprocessor`.

The following flow works for me. Please help test.

```
// start server in detached mode
cartesi rollups start -d -v

// check if server is running
cartesi rollups status

// create a new deroll v2 project
npm init @deroll/app@2.0.0-alpha.0

// inside the dapp...
cartesi build
cartesi rollups deploy

// send an input
cartesi send

// stop the server
cartesi rollups stop
```


apps/cli/src/node/docker-compose-anvil.yaml
```diff
@@ -19,27 +19,29 @@ services:
...
-                "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
-                "0x4C11C7F82D6D56a726f9B53dd99af031AFd86BB6",
-                "deployContracts(address,address,bytes32,bytes32)",
-                "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
-                "720",
-                "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266",
-                "0x0000000000000000000000000000000000000000000000000000000000000000",
-                "0x0000000000000000000000000000000000000000000000000000000000000000",
-            ]
+            - /bin/bash
```
tuler: I usually don't like to use `bash` as the command, and instead use the actual process we need to start.
Is this because of env variables?
I understand as well that we need the command stdout to go to a file, in the host `.cartesi` directory. How are we doing that?

endersonmaia: Yeah, it was a way to debug with -x

About saving do the host, we'd need to mount it into the container and redirect to a file.

Or we'd need to save to a named volume and copy on another process with docker cp.

tuler: mounting the volume is easy, but how to redirect to a file? Does the following work??
```
- ">"
- "filename"
```

tuler: Support for environment variables will be added to the cartesi-rollups-cli

tuler:
```
$ cartesi rollups start --help
Usage: cartesi rollups start [options]

Start a local rollups node environment.

Options:
  --block-time <number>     interval between blocks (in seconds) (default: 5)
  --default-block <string>  default block to be used when fetching new blocks. (choices: "latest", "safe", "pending", "finalized", default: "finalized")
  --cpus <number>           number of cpu limits for the rollups-node
  --memory <number>         memory limit for the rollups-node in MB
  --project-name <string>   name of environment (default: "cartesi-rollups")
  --services <string>       optional services to start, comma separated list from [bundler, explorer, graphql, paymaster] (default: [])
  -p, --port <number>       port to listen on (default: 8080)
  -d, --detach              run in detached mode (default: false)
  --dry-run                 show the docker compose configuration (default: false)
  -v, --verbose             verbose output (default: false)
  -h, --help                display help for command
```

tuler:
```
$ cartesi rollups stop --help 
Usage: cartesi rollups stop [options]

Stop a local rollups node environment.

Options:
  --project-name <project-name>  Project name (default: "cartesi-rollups")
  -h, --help                     display help for command
```

tuler:
```
$ cartesi rollups stop --help 
Usage: cartesi rollups stop [options]

Stop a local rollups node environment.

Options:
  --project-name <project-name>  Project name (default: "cartesi-rollups")
  -h, --help                     display help for command
➜  test2 cartesi-dev rollups status --help
Usage: cartesi rollups status [options]

Shows the status of a local rollups node environment.

Options:
  --project-name <string>  name of environment (default: "cartesi-rollups")
  --json                   output in JSON format
  -h, --help               display help for command
```

tuler:
```
$ cartesi rollups deploy --help
Usage: cartesi rollups deploy [options]

Deploy a rollups application to a rollups node.

Options:
  --chain-id <id>                Chain ID (default: 31337)
  --rpc-url <url>                RPC URL
  --mnemonic <phrase>            Mnemonic passphrase
  --mnemonic-index <index>       Mnemonic account index (default: 0)
  --project-name <string>        name of environment (default: "cartesi-rollups")
  --name <string>                application name
  --authority-owner <address>    authority owner
  --application-owner <address>  application owner
  --epoch-length <number>        length of an epoch (in blocks) (default: 720)
  --salt <hash>                  salt for deployment (default: "0x0000000000000000000000000000000000000000000000000000000000000000")
  --json                         output in JSON format
  -h, --help                     display help for command
```

apps/cli/src/commands/rollups/deploy.ts
```diff
@@ -0,0 +1,350 @@
...
+ * @returns path to the snapshot in the rollups node
+ */
+const publishMachine = async (options: {
+    progress: Ora;
+    projectName: string;
+    templateHash: Hash;
+}): Promise<string> => {
+    const { progress, projectName, templateHash } = options;
+    const snapshotPath = getContextPath("image");
+    const containerSnapshotPath = `/var/lib/cartesi-rollups-node/snapshots/${templateHash}/`;
```
endersonmaia, line 155: If this path is hardcoded here, we should make sure that ovewriting `CARTESI_SNAPSHOT_DIR` won't break anything.

apps/cli/src/commands/rollups/start.ts
```diff
@@ -0,0 +1,196 @@
...
+                        stdio: "inherit",
+                    });
+                    return;
+                }
+
+                // run compose environment
+                await execa("docker", [...compose_args, "up", ...up_args], {
+                    env,
+                    stdio: "inherit",
+                });
```
endersonmaia, line 173: We should avoid breaking thing if a second `cartesi rollups start` is called.

Print a message that there's a cartesi-rollups environment already running is enough for now.

apps/cli/src/commands/send.ts
```diff
@@ -66,6 +65,13 @@ export const addCommonOptions = (command: Command) => {
         );
 };
 
+export const addCommonOptions = (command: Command) => {
+    return addConnectOptions(command).option(
+        "--dapp <address>",
+        "Application address",
+    );
+};
+
```
endersonmaia: When testing the `send` command having two dapps deployed, the interactive cli ignores the `--dapp` defined and suggests the first dapp in the list.


Screenshot from 2025-03-06 10-50-41


Screenshot from 2025-03-06 10-51-16

apps/cli/src/compose/rollups/docker-compose-envfile.yaml
```diff
@@ -1,4 +1,4 @@
 services:
-    validator:
+    rollups-node:
```
endersonmaia: maybe just node so we get a container named `cartesi-rollups-node` instead of `cartesi-rollups-rollups-node`.

Alsso calling  `docker compose -p cartesi-rollups run -ti node /bin/bash` looks better, if needed.

No hard opinion, just aesthetics.

apps/cli/src/compose/rollups/docker-compose-espresso.yaml
```diff
@@ -0,0 +1,59 @@
...
+            PGHOST: ${PGHOST:-database}
+            PGPORT: ${PGPORT:-5432}
+            PGUSER: ${PGUSER:-postgres}
+            PGPASSWORD: ${PGPASSWORD:-password}
+            PGDATABASE: ${PGDATABASE:-postgres}
+
+    rollups-node:
+        environment:
+            CARTESI_FEATURE_INPUT_READER_ENABLED: false
+    espresso:
```
endersonmaia:
```suggestion
            CARTESI_FEATURE_INPUT_READER_ENABLED: false

    espresso:
```

apps/cli/src/compose/rollups/docker-compose-espresso.yaml
```diff
@@ -0,0 +1,59 @@
...
+        image: debian:bookworm-slim
+        environment:
+            PROMPT_TXT_07_ESPRESSO: "Espresso running at http://localhost:${CARTESI_LISTEN_PORT}/espresso/"
+
+    proxy:
+        volumes:
+            - ./proxy/espresso-dev.yaml:/etc/traefik/conf.d/espresso-dev.yaml
+            - ./proxy/espresso-sequencer.yaml:/etc/traefik/conf.d/espresso-sequencer.yaml
+            - ./proxy/espresso-builder.yaml:/etc/traefik/conf.d/espresso-builder.yaml
+            - ./proxy/espresso-prover.yaml:/etc/traefik/conf.d/espresso-prover.yaml
```
endersonmaia: Maybe we won't even expose any of those, but couldn't we just use a single `espresso.yaml` file?

A good convention could be `./proxy/<service>.yaml`

apps/cli/src/compose/rollups/docker-compose-node.yaml
```diff
@@ -0,0 +1,29 @@
...
+        command: ["cartesi-rollups-node"]
+        env_file:
+            - ${CARTESI_BIN_PATH}/compose/rollups/default.env
+
+    proxy:
+        depends_on:
+            rollups-node:
+                condition: service_healthy
+        volumes:
+            - ./proxy/rollups-node.yaml:/etc/traefik/conf.d/rollups-node.yaml
```
endersonmaia, line 37:
```suggestion
            - ./proxy/rollups-node.yaml:/etc/traefik/conf.d/rollups-node.yaml

```

apps/cli/src/compose/rollups/proxy/rollups-node.yaml
```diff
@@ -0,0 +1,10 @@
+http:
+    routers:
+        inspect_server:
+            rule: "PathPrefix(`/inspect`)"
+            service: inspect_server
+    services:
+        inspect_server:
+            loadBalancer:
+                servers:
+                    - url: "http://rollups-node:10012/inspect"
```
endersonmaia, line 13: This is why I didn't use file and used inline content in YAML files, so we could replace values with environment variables, like this `10012` for `CARTESI_INSPECT_PORT`, for ex.

apps/cli/src/compose/rollups/default.env
```diff
@@ -0,0 +1,42 @@
...
+
+# auth
+CARTESI_AUTH_MNEMONIC="${CARTESI_AUTH_MNEMONIC:-test test test test test test test test test test test junk}"
+
+# postgres
+CARTESI_DATABASE_CONNECTION="${CARTESI_DATABASE_CONNECTION:-postgres://postgres:password@database:5432/postgres?sslmode=disable}"
+
+# http
+CARTESI_INSPECT_PORT=${CARTESI_INSPECT_PORT:-10012}
+CARTESI_INSPECT_ADDRESS=${CARTESI_INSPECT_ADDRESS:-0.0.0.0:$CARTESI_INSPECT_PORT}
```
endersonmaia: This will be useless if we use the `./proxy/*.yaml` files, unless we find a way to replace it in the files.

apps/cli/src/compose/rollups/default.env
```diff
@@ -0,0 +1,42 @@
...
+CARTESI_BLOCKCHAIN_DEFAULT_BLOCK="${CARTESI_BLOCKCHAIN_DEFAULT_BLOCK:-finalized}"
+
+# contracts
+CARTESI_CONTRACTS_INPUT_BOX_ADDRESS="${CARTESI_CONTRACTS_INPUT_BOX_ADDRESS:-0x593E5BCf894D6829Dd26D0810DA7F064406aebB6}"
+CARTESI_CONTRACTS_INPUT_BOX_DEPLOYMENT_BLOCK_NUMBER="${CARTESI_CONTRACTS_INPUT_BOX_DEPLOYMENT_BLOCK_NUMBER:-1}"
+CARTESI_CONTRACTS_APPLICATION_FACTORY_ADDRESS="${CARTESI_CONTRACTS_APPLICATION_FACTORY_ADDRESS:-0xd7d4d184b82b1a4e08f304DDaB0A2A7a301C2620}"
+CARTESI_CONTRACTS_AUTHORITY_FACTORY_ADDRESS="${CARTESI_CONTRACTS_AUTHORITY_FACTORY_ADDRESS:-0xB897F7Fe78f220aE34B7FA9493092701a873Ed45}"
+
+# snapshot
+CARTESI_SNAPSHOT_DIR="${CARTESI_SNAPSHOT_DIR:-/var/lib/cartesi-rollups-node/snapshots}"
```
endersonmaia: Like I commented elsewhere, this won't me useful if we are hard-coding it, maybe we should review this environment variable config approach.

apps/cli/src/base.ts
endersonmaia, line 1: I thins multu dapp approach, `cartesi address-book` showing the application address doesn't make much sense, right?

endersonmaia (commented):
Since this PR is mine, I can't request changes.

apps/cli/src/commands/rollups/deploy.ts
tuler: I think a lot of the env variables that we used to allow redefinition we should not.

apps/cli/src/compose/rollups/docker-compose-envfile.yaml
tuler: Yeah, I thought `node` would be too generic.

apps/cli/src/compose/rollups/docker-compose-espresso.yaml
tuler: I don't know if Traefik config file support multiple in a single file, does it?

tuler: Regarding exposing it, I think in a future alpha, maybe the one which includes the espresso reader, we can stop exposing it.

apps/cli/src/compose/rollups/proxy/rollups-node.yaml
tuler: I understand, but I think a lot of env vars we should stop using, and hard code for simplicity.

apps/cli/src/compose/rollups/default.env
tuler: I think we need to clean the env and leave just what really makes sense to be overridden.

apps/cli/src/base.ts
tuler: Yeah, maybe it still makes sense.
The idea is that a project has only 1 machine snapshot, with a hash.
If you run `address-book` from the project directory it tries to find a deployment for that hash.

apps/cli/src/commands/send.ts
tuler: Fixed

apps/cli/src/compose/rollups/docker-compose-envfile.yaml
endersonmaia: > Yeah, I thought node would be too generic.

I had the same thought, and that's why I went with `validator`.

tuler: I don't like validator because it's also a reader.
The validator vs reader thing is still a bit clouded, with the prt-node effort in parallel.
That's why I wanted to avoid the `validator` term.

apps/cli/src/compose/rollups/docker-compose-espresso.yaml
tuler: Done unifying proxy config

apps/cli/src/compose/rollups/docker-compose-node.yaml
```diff
@@ -0,0 +1,39 @@
...
+    rollups-node-migration:
+        image: cartesi/rollups-node:devel
+        command: ["cartesi-rollups-cli", "db", "upgrade"]
+        depends_on:
+            database:
+                condition: service_healthy
+        restart: "no"
+        env_file:
+            - ${CARTESI_BIN_PATH}/compose/rollups/default.env
+    rollups-node:
```
endersonmaia:
```suggestion
            - ${CARTESI_BIN_PATH}/compose/rollups/default.env

    rollups-node:
```

endersonmaia (commented):
The `send` behavior is still weird.

Even if I define the `--dapp` argument, it will override with the address of the dapp from the directory I'm calling the command from.

`--dapp` should take precedence

apps/cli/src/commands/rollups/start.ts
```diff
@@ -0,0 +1,196 @@
...
+import chalk from "chalk";
+import { execa } from "execa";
+import path from "path";
+
+const commaSeparatedList = (value: string, _previous: string[]) =>
+    value.split(",");
+
+const availableServices = [
+    "bundler",
+    "explorer",
```
endersonmaia, line 13:
```suggestion
    "bundler",
    "espresso",
    "explorer",
```

tuler: If we agree with `--sequencer espresso`, then `espresso` would not be enabled by `--services`, so it should not be part of this list.

tuler:
> `--dapp` should take precedence

I think this is commander bug. Opened an issue at https://github.com/tj/commander.js/issues/2335

tuler:
> I think this is commander bug. Opened an issue at [tj/commander.js#2335](https://github.com/tj/commander.js/issues/2335)

Problem solved.

ps: there are two ways of structuring commands and subcommands in commander.

1) `.command(name).`
2) `.addCommand(new Command(name))`

I was using (1), but switched to (2).

tuler:
Need to check why the graphql container is continuously printing:
```
WRN No raw input references found
```

apps/cli/src/commands/send/generic.ts
```diff
@@ -114,12 +112,17 @@ export const registerGenericCommand = (program: Command) => {
             ).choices(["hex", "string", "abi"]),
         )
         .option("--input-abi-params <input-abi-params>", "input abi params")
-        .action(async (options) => {
+        .action(async (options, command) => {
+            const sendOptions = command.optsWithGlobals();
+            console.log(sendOptions, options);
```
endersonmaia:
```suggestion
```

endersonmaia (commented):
`... send --dapp=` is working as expected

endersonmaia:
@tuler I think we're good to go

tuler:
Regarding the start + start issue, I was thinking of an elegant way to handle it without checking first if it's running.

What happens now is that docker compose up, when executed again, "reapplies" the services, which is usually harmless. But the one-shot containers get reexecutes, and the one which creates the graphql database fails because the database already exists.

The solutions to create only if exists are all ugly and not elegant.

endersonmaia:
> @tuler I think we're good to go

except that we need a rollups-node v2 release :melting_face:

tuler:
Applied the idea of https://github.com/cartesi/cli/pull/151 but without the extends, just using an env var for the SDK image name.

apps/cli/src/commands/rollups/start.ts
```diff
@@ -0,0 +1,194 @@
...
+            const env: NodeJS.ProcessEnv = {
+                ANVIL_VERBOSITY: verbose ? "--steps-tracing" : "--silent",
+                BLOCK_TIME: blockTime.toString(),
+                CARTESI_BLOCKCHAIN_DEFAULT_BLOCK: defaultBlock,
+                CARTESI_LOG_LEVEL: verbose ? "info" : "error",
+                CARTESI_BIN_PATH: binPath,
+                CARTESI_LISTEN_PORT: listenPort.toString(),
+                CARTESI_ROLLUPS_NODE_CPUS: cpus?.toString(),
+                CARTESI_ROLLUPS_NODE_MEMORY: memory?.toString(),
+                SDK_IMAGE: DEFAULT_SDK,
```
endersonmaia:
```suggestion
                CARTESI_SDK_IMAGE: DEFAULT_SDK,
```

apps/cli/src/compose/rollups/proxy/rollups-node.yaml
```diff
@@ -0,0 +1,17 @@
...
+            service: rpc_server
+    services:
+        inspect_server:
+            loadBalancer:
+                servers:
+                    - url: "http://rollups-node:10012/inspect"
+        rpc_server:
+            loadBalancer:
+                servers:
+                    - url: "http://rollups-node:10011/inspect"
```
endersonmaia, line 17:
```suggestion
                    - url: "http://rollups-node:10011/rpc"
```

```suggestion
                    - url: "http://rollups-node:10011/json-rpc"
```

tuler: My experience from other projects is that `rpc` is more common than `json-rpc` for endpoints

endersonmaia: Do you think we should use a `/rollups/inspect` adn `/rollups/rpc` to hide/abastract the port?

Or even `/cartesi/node/rpc`, `/cartesi/node/inspect`

tuler: I'd leave as it is for simplicity. Node is the main service of the environment, it's why the environment exist, it deserves the "root path"

tuler (approved):

PR #121, opened 2024-10-30
Bump sdk

tuler:


apps/cli/src/node/docker-compose-anvil.yaml
```diff
@@ -1,6 +1,6 @@
 services:
     anvil:
-        image: cartesi/sdk:0.12.0-alpha.0
+        image: cartesi/sdk:0.12.0-alpha.1
```
endersonmaia, line 3: What if we put this value in a environment variable in a single place?

https://github.com/cartesi/cli/blob/feature/bump-sdk/apps/cli/src/commands/run.ts#L102

At least for the `cartesi run` it would be useful.

endersonmaia (approved):
Approving with a small suggestion, won't block for it.

PR #129, opened 2024-11-08
Set default ram-image to always be /usr/share/cartesi-machine/images/linux.bin

tuler:
This sets the default ram-image path to always be `/usr/share/cartesi-machine/images/linux.bin`, regardless the system os.
This is the path inside the SDK image that contains the `linux.bin` as well, so it works if the user does not have `cartesi-machine` installed on the host, which falls back to executing inside the SDK image container.
If the user does have `cartesi-machine` installed, and the `linux.bin` image is located somewhere else, he will need to define the path in the `cartesi.toml` config file.

### DIFF
--- a/.changeset/perfect-news-compare.md
+++ b/.changeset/perfect-news-compare.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": major
+---
+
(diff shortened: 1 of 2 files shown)


endersonmaia (approved):
Tested on Ubuntu Linux 24.04, with and without `cartesi-machine` available in the host.

PR #150, opened 2025-02-03
feat(cli): add espresso-reader

endersonmaia:
You can test this with:

```
pnpm build --filter @cartesi/cli
alias cartesi-dev=$PWD/apps/cli/dist/index.js
cartesi-dev create --branch=prerelease/sdk-12 --template=go my-go-dapp-v2 
cd my-go-dapp-v2
cartesi-dev build
cartesi-dev rollups start --services espresso,graphql
cartesi-dev rollups deploy
```

The `rollups-espresso-reader` endpoints should be available at:

- http://localhost:8080/espresso/reader/nonce
- http://localhost:8080/espresso/reader/submit

### DIFF
--- a/apps/cli/src/compose/rollups/docker-compose-espresso.yaml
+++ b/apps/cli/src/compose/rollups/docker-compose-espresso.yaml
@@ -6,15 +6,11 @@ services:
             database:
                 condition: service_healthy
         environment:
-            PGHOST: ${PGHOST:-database}
-            PGPORT: ${PGPORT:-5432}
-            PGUSER: ${PGUSER:-postgres}
-            PGPASSWORD: ${PGPASSWORD:-password}
-            PGDATABASE: ${PGDATABASE:-postgres}
-
-    rollups-node:
-        environment:
-            CARTESI_FEATURE_INPUT_READER_ENABLED: false
+            PGHOST: database
+            PGPORT: 5432
+            PGUSER: postgres
+            PGPASSWORD: password
+            PGDATABASE: postgres
 
     espresso:
         image: ${CARTESI_SDK_IMAGE}
@@ -32,7 +28,7 @@ services:
         depends_on:
             espresso_database_creator:
                 condition: service_completed_successfully
-            database:
+            anvil:
                 condition: service_healthy
         environment:
             ESPRESSO_SEQUENCER_L1_PROVIDER: ${CARTESI_BLOCKCHAIN_HTTP_ENDPOINT:-http://anvil:8545}
@@ -46,6 +42,8 @@ services:
             ESPRESSO_SEQUENCER_POSTGRES_PASSWORD: password
             ESPRESSO_SEQUENCER_POSTGRES_DATABASE: sequencer
             ESPRESSO_SEQUENCER_ETH_MNEMONIC: ${CARTESI_AUTH_MNEMONIC:-test test test test test test test test test test test junk}
+            ESPRESSO_SEQUENCER_L1_POLLING_INTERVAL: "1s"
+            ESPRESSO_STATE_PROVER_UPDATE_INTERVAL: "1s"
         healthcheck:
             test:
                 [
@@ -60,9 +58,46 @@ services:
             timeout: 1s
             retries: 5
 
+    rollups-node:
+        environment:
+            CARTESI_FEATURE_INPUT_READER_ENABLED: false
+
+    espresso_reader_migration:
+        image: ${CARTESI_SDK_IMAGE}
+        command:
+            - migrate
+            - -source
+            - file:///usr/share/cartesi/rollups-espresso-reader/migrations
+            - -database
+            - postgres://postgres:password@database:5432/postgres?sslmode=disable&x-migrations-table=espresso_schema_migrations
+            - up
+        depends_on:
+            rollups-node-migration:
+                condition: service_completed_successfully
+
+    espresso_reader:
+        image: ${CARTESI_SDK_IMAGE}
+        command: ["cartesi-rollups-espresso-reader"]
+        env_file:
+            - ${CARTESI_BIN_PATH}/compose/rollups/default.env
+        ports:
+            - 8081
+        depends_on:
+            espresso_reader_migration:
+                condition: service_completed_successfully
+            espresso:
+                condition: service_healthy
+        environment:
+            CARTESI_POSTGRES_ENDPOINT: postgres://postgres:password@database:5432/postgres?sslmode=disable
+            ESPRESSO_SERVICE_ENDPOINT: ":8081"
+            ESPRESSO_BASE_URL: http://espresso:8770
+            ESPRESSO_NAMESPACE: 51025
+            ESPRESSO_STARTING_BLOCK: 101
+
     proxy:
         depends_on:
             espresso:
                 condition: service_healthy
         volumes:
             - ./proxy/espresso.yaml:/etc/traefik/conf.d/espresso.yaml
+            - ./proxy/espresso-reader.yaml:/etc/traefik/conf.d/espresso-reader.yaml
--- a/.changeset/tasty-chairs-doubt.md
+++ b/.changeset/tasty-chairs-doubt.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+add rollups-espresso-reader service
--- a/apps/cli/src/compose/rollups/proxy/espresso-reader.yaml
+++ b/apps/cli/src/compose/rollups/proxy/espresso-reader.yaml
@@ -1,0 +1,17 @@
+http:
+    routers:
+        espresso-reader:
+            rule: "PathPrefix(`/espresso/reader`)"
+            middlewares:
+                - "remove-espresso-reader-prefix"
+            service: espresso-reader
+    middlewares:
+        remove-espresso-reader-prefix:
+            replacePathRegex:
+                regex: "^/espresso/reader/(.*)"
+                replacement: "/$1"
+    services:
+        espresso-reader:
+            loadBalancer:
+                servers:
+                    - url: "http://espresso_reader:8081"


apps/cli/src/node/docker-compose-espresso.yaml
```diff
@@ -45,6 +91,11 @@ services:
             ESPRESSO_SEQUENCER_POSTGRES_PASSWORD: password
             ESPRESSO_SEQUENCER_POSTGRES_DATABASE: sequencer
             ESPRESSO_SEQUENCER_ETH_MNEMONIC: ${CARTESI_AUTH_MNEMONIC:-test test test test test test test test test test test junk}
+            ESPRESSO_STATE_PROVER_UPDATE_INTERVAL: "1s"
+            ESPRESSO_STATE_PROVER_RETRY_INTERVAL: "500ms"
```
miltonjonat: what are these ESPRESSO_STATE_PROVER variables?

tuler: According to the docs at https://github.com/EspressoSystems/espresso-sequencer/blob/main/doc/espresso-dev-node.md

ESPRESSO_STATE_PROVER_UPDATE_INTERVAL: The frequency of updating the light client state
ESPRESSO_STATE_PROVER_RETRY_INTERVAL: Interval between retries if a state update fails.

miltonjonat: Oh right. Sorry, I completely misread this and thought these were configs for our reader, not Espresso itself!

apps/cli/src/node/docker-compose-espresso.yaml
```diff
@@ -15,9 +15,53 @@ services:
...
+        env_file:
+            - ${CARTESI_BIN_PATH}/node/default.env
+        depends_on:
+            database:
+                condition: service_healthy
+        environment:
+            DATABASE_URL: postgres://postgres:password@database:5432/espresso_reader
+
+    espresso_reader:
+        image: cartesi/sdk:devel
```
miltonjonat:
```suggestion
        image: cartesi/sdk:0.12.0-alpha.5
```

miltonjonat:
Instructions fixing typos :)
```
pnpm build --filter @cartesi/cli
alias cartesi-dev=$PWD/apps/cli/bin/run.js
cartesi-dev create --branch=prerelease/sdk-12 --template=go my-go-dapp-v2 
cd my-go-dapp-v2
cartesi-dev build
cartesi-dev run --enable-espresso
```

miltonjonat (commented):
I couldn't access the espresso endpoint on http://localhost:8080/espresso, only directly on http://localhost:8770. Is that expected?

endersonmaia:
> I couldn't access the espresso endpoint on http://localhost:8080/espresso, only directly on http://localhost:8770. Is that expected?

It's missing some instructions. We expose 4 espresso API endpoints via names PATHs, they are:

- `http://localhost:8080/espresso/sequencer`
- `http://localhost:8080/espresso/builder`
- `http://localhost:8080/espresso/prover`
- `http://localhost:8080/espresso/dev`

So if you wanna call the `/v0/status/block-height` you should call

`curl http://localhost:8080/espresso/sequencer/v0/status/block-height`

apps/cli/src/node/docker-compose-espresso.yaml
```diff
@@ -15,9 +15,53 @@ services:
...
+                condition: service_completed_successfully
+            espresso:
+                condition: service_healthy
+        env_file:
+            - ${CARTESI_BIN_PATH}/node/default.env
+        environment:
+            DATABASE_URL: postgres://postgres:password@database:5432/espresso_reader
+            ESPRESSO_BASE_URL: http://espresso:8770
+            ESPRESSO_NAMESPACE: 51025
+            ESPRESSO_STARTING_BLOCK: 101
```
miltonjonat: The rollups-espresso-reader does a weird thing: it publishes two endpoints `/nonce` and `/submit`  on a port configurable with env ESPRESSO_SERVICE_ENDPOINT (default I think is 8080). We'd need to expose these too.

endersonmaia: What do you think about this format?

- `/espresso/reader/nonce`
- `/espresso/reader/submit`

If the developer interaction if gonna be focused on those endpoints, we could even hide everything from `espresso-dev-node` and only publish those at `/espresso/submit` and `/espresso/nonce`, also.

apps/cli/src/node/docker-compose-espresso.yaml
endersonmaia: Already bumped to new sdk:0.12.0-alpha.6 that already containes espresso-dev-node with ;latest `patch5` tag.

apps/cli/src/node/docker-compose-espresso.yaml
endersonmaia: @miltonjonat I sent a fixup where you can access those endpoints at:

 - `http://localhost:8080/espresso/reader/nonce`
- `http://localhost:8080/espresso/reader/submit`

Please, let me know if it works!

miltonjonat: So, these endpoints are currently conveniences for the Espresso integration, and strictly speaking they could (or should?) be part of another component separate from the reader itself. So, no, I wouldn't use "reader" there.

Frankly, I think it's also quite odd to put it under "espresso" in this context, because that "espresso" there is meant to be the (local) Espresso Network, while these are endpoints on the Node itself.

Tbh, the idea here was also that the Node would provide `/nonce` and `/submit` functionalities as something more universal (i.e., to also be used when integrated with Avail, Celestia, etc). And that clients wouldn't even know which alt-DA was being used by the app/node.

**TL;DR**, I'd use http://localhost:8080/nonce and http://localhost:8080/submit directly; *or* we could come up with a nice namespace for convenience endpoints of this kind (maybe http://localhost:8080/tx/nonce?)

tuler: I understand there is no unified way to send a transaction. anvil expects ethereum transactions as specified by [eth_sendTransaction](https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_sendtransaction/), espresso expects a [transaction](https://docs.espressosys.com/network/api-reference/sequencer-api#transaction) object, avail has its own [spec](https://docs.availproject.org/api-reference/avail-node-api/da-submit-data). 

So having a single `/nonce` and `/submit` at the root of the service, regardless the configured DA, with various formats will be extremely confusing.

tuler: If the only APIs the application developer will use in case he chooses espresso as DA are the `/nonce` and `/submit`, I wouldn't even expose the espresso-dev-node APIs to the host, as they would be used internally (inside docker network) by the `espresso-reader`.

The application developer APIs would be at http://localhost:8080/espresso/nonce and http://localhost:8080/espresso/submit

Does the developer need direct access to espresso-dev-node?

miltonjonat: I agree, I don't expect the developer to have to access the Espresso Network directly (espresso-dev-node or Espresso Decaf Testnet in case he's using that). It's just part of the environment, may be somewhat useful to see how things are working if you want to debug deeper, but definitely not necessary.

endersonmaia: I'm exposing those espresso-dev-node ports to the host to test if thins are working, but they should be removed on the final version of this PR togethers with the `/espresso` endpoints for the proxy.

About the `/nonce` and `/submit` endpoints from the espresso-reader, I need a definition so I can apply in this PR.

miltonjonat: > I understand there is no unified way to send a transaction. anvil expects ethereum transactions as specified by [eth_sendTransaction](https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_sendtransaction/), espresso expects a [transaction](https://docs.espressosys.com/network/api-reference/sequencer-api#transaction) object, avail has its own [spec](https://docs.availproject.org/api-reference/avail-node-api/da-submit-data).
> 
> So having a single `/nonce` and `/submit` at the root of the service, regardless the configured DA, with various formats will be extremely confusing.

I see this differently. With Paio, we standardized an [EIP-712 structure to be used when submitting tx's](https://github.com/cartesi/rollups-espresso-reader/blob/main/docs/development.md#signing-a-transaction). This structure includes the target chainId and app address, nonce, and envisions payment info too. Avail and Celestia integrations would use this via Paio, and we adopted it for the Espresso integration as well (Espresso itself doesn't have any spec on the payload you submit, which doesn't even need to be signed atm). The idea is that you would use `/nonce` and `/submit` for all L2 tx submissions (i.e., that use Paio / alt-DA), while of course clients would still send direct L1 Eth tx's via the InputBox for L1->L2 messages.

Now, another interesting discussion would be to provide a standard Eth JSON-RPC API on the Node for these L2 tx's. I'd like that, but I don't know how viable it is. It would be part of that redesign discussion for all this architecture.

miltonjonat: So, one path what would be very adherent to the current architecture would be something like `L2tx`, leading to http://localhost:8080/L2tx/nonce and http://localhost:8080/L2tx/submit (I suggested `tx` above, but I admit that `L2tx` would be more precise)

tuler: How are you going to use a single format for DAs with different requirements?
I.e. how do you define the namespace in case of Espresso?

miltonjonat: In the current design, the client/user does not inform any DA-specific parameters in the payload it submits. Whatever is required to submit to the alt-DA configured for the application should be given by DA-specific parameters defined upon deployment of the app, and stored as on-chain app configuration. The Node will fetch that from the application contract (it's currently defined as an ABI encoded method+parameters), and thus know how to handle tx submission when `/submit` is called.

miltonjonat:
I tried it out, had some issues, and couldn't run in the end.
1. `./apps/cli/bin/run.js` is not being generated by `pnpm build`
2. If I go to that directory and run `pnpm build`, it generates output in the `dist` folder. It seems to work if I use `alias cartesi-dev=$PWD/apps/cli/dist/index.js`; using `cartesi-dev build` works
3. `cartesi-dev run --enable-espresso` fails, with message `service "espresso_reader_migration" didn't complete successfully: exit 1`. Full log (with verbose): [cartesi-dev.run.log.zip](https://github.com/user-attachments/files/19061500/cartesi-dev.run.log.zip)

endersonmaia:
> I tried it out, had some issues, and couldn't run in the end.
> 
> 1. `./apps/cli/bin/run.js` is not being generated by `pnpm build`
> 2. If I go to that directory and run `pnpm build`, it generates output in the `dist` folder. It seems to work if I use `alias cartesi-dev=$PWD/apps/cli/dist/index.js`; using `cartesi-dev build` works
> 3. `cartesi-dev run --enable-espresso` fails, with message `service "espresso_reader_migration" didn't complete successfully: exit 1`. Full log (with verbose): [cartesi-dev.run.log.zip](https://github.com/user-attachments/files/19061500/cartesi-dev.run.log.zip)

The failure was `error: migration failed: type "ethereum_address" does not exist`.

The migration code doesn't mention the creation of this `ethereum_address` type.

https://github.com/cartesi/rollups-espresso-reader/blob/v0.2.1-node-20250128/internal/repository/postgres/schema/migrations/000001_espresso_db.up.sql

miltonjonat:
That's part of the Node migration, which has to be executed before the Espresso Reader's one.
https://github.com/cartesi/rollups-node//blob/feature/new-build-20250128/internal/repository/postgres/schema/migrations/000001_create_initial_schema.up.sql#L4

endersonmaia:
> That's part of the Node migration, which has to be executed before the Espresso Reader's one. https://github.com/cartesi/rollups-node//blob/feature/new-build-20250128/internal/repository/postgres/schema/migrations/000001_create_initial_schema.up.sql#L4

Sent a fixup for that. [0fb7ac6](https://github.com/cartesi/cli/pull/150/commits/0fb7ac6ea701737a1798d8aeb3325e483f199a7f)

I'll have to rebase this on top of `feature/cli-run-v2`, it has a lot of braking changes, then I'll provide a cleaner commit history so we can go back to testing this again.

miltonjonat:
@endersonmaia cool new instructions! I like the `--services espresso,graphql` thing.
Unfortunately, it didn't run: migrations failed again. It seems that the database is dying for some reason, even when running without any extra services.

- Log with both espresso and graphql services: [start-espresso-graphql.log.gz](https://github.com/user-attachments/files/19171698/start-espresso-graphql.log.gz)
- Log with only espresso service: [start-espresso.log.gz](https://github.com/user-attachments/files/19171699/start-espresso.log.gz)
- Log without any extra service: [start-no-services.log.gz](https://github.com/user-attachments/files/19171700/start-no-services.log.gz)

tuler:
> * Log without any extra service: [start-no-services.log.gz](https://github.com/user-attachments/files/19171700/start-no-services.log.gz)

Strange that it is reporting a database url as `postgres://postgres:password@localhost:5432/rollupsdb?sslmode=disable`, with database name as `rollupsdb`. It should be `postgres`.

Did you change any of that?

miltonjonat:
> > * Log without any extra service: [start-no-services.log.gz](https://github.com/user-attachments/files/19171700/start-no-services.log.gz)
> 
> Strange that it is reporting a database url as `postgres://postgres:password@localhost:5432/rollupsdb?sslmode=disable`, with database name as `rollupsdb`. It should be `postgres`.
> 
> Did you change any of that?

I didn't change anything, but AFAIK in Node v2 the database is always called `rollupsdb` (e.g., https://github.com/cartesi/rollups-node/blob/feature/new-build/Makefile#L82)
Btw, which Node branch/tag are you guys using? The Espresso Reader is currently on https://github.com/cartesi/rollups-node/releases/tag/v2.0.0-dev-20250128

tuler:
> Btw, which Node branch/tag are you guys using?

We are testing locally with branch `feature/new-build` while we wait for a tag.

tuler:
> AFAIK in Node v2 the database is always called `rollupsdb

We define the env var https://github.com/cartesi/cli/blob/8c2ba890dcd66009ceb07545bdb5f1df2d0870ad/apps/cli/src/compose/rollups/default.env#L35

miltonjonat:
Right.. maybe it's not going through, or not being respected by the Node code..?

endersonmaia:
`rollups-node` v2 [changed](https://github.com/cartesi/rollups-node/commit/81b533fa44cb478cf0309624598176b0e90353ae) the ENV config from `CARTESI_POSTGRES_ENDPOINT` to `CARTESI_DATABASE_CONNECTION`

I'm already using `CARTESI_ENDPOINT_CONNECTION` where @tuler pointed out in his last comment.

For espresso-reader, I'm [using](https://github.com/cartesi/cli/blob/f6996c44434770157a257ed19b86a6a3ea3777be/apps/cli/src/compose/rollups/docker-compose-espresso.yaml#L46) `CARTESI_POSTGRES_ENDPOINT` still.

endersonmaia:
@miltonjonat my suggestion is that you go to the cartesi/rollups-node local repository at the `feature/new-build` branch and run `make image`.

Then you'll have a `cartesi/rollups-node:devel` available locally and your tests may run just fine. 

Let me know if anything changes.

tuler:
Time to rebase this one.

apps/cli/src/compose/rollups/docker-compose-espresso.yaml
```diff
@@ -66,3 +101,4 @@ services:
                 condition: service_healthy
         volumes:
             - ./proxy/espresso.yaml:/etc/traefik/conf.d/espresso.yaml
+            - ./proxy/espresso-reader.yaml:/etc/traefik/conf.d/expresso-reader.yaml
```
tuler:
```suggestion
            - ./proxy/espresso-reader.yaml:/etc/traefik/conf.d/espresso-reader.yaml
```

apps/cli/src/compose/rollups/docker-compose-espresso.yaml
```diff
@@ -16,6 +16,39 @@ services:
         environment:
             CARTESI_FEATURE_INPUT_READER_ENABLED: false
 
+    espresso_reader_migration:
+        image: cartesi/sdk:0.12.0-alpha.8
```
tuler:
```suggestion
        image: ${CARTESI_SDK_IMAGE}
```

apps/cli/src/compose/rollups/docker-compose-espresso.yaml
```diff
@@ -16,6 +16,39 @@ services:
...
+        command:
+            - /bin/bash
```
tuler: I don't think we need bash

apps/cli/src/compose/rollups/docker-compose-espresso.yaml
```diff
@@ -16,6 +16,39 @@ services:
...
+                migrate                                                                                                                         \
+                -source file:///usr/share/cartesi/rollups-espresso-reader/migrations                                                            \
+                -database "postgres://postgres:password@database:5432/postgres?sslmode=disable&x-migrations-table=espresso_schema_migrations"   \
+                up
+        depends_on:
+            rollups-node-migration:
+                condition: service_completed_successfully
+
+    espresso_reader:
+        image: cartesi/sdk:0.12.0-alpha.8
```
tuler:
```suggestion
        image: ${CARTESI_SDK_IMAGE}
```

apps/cli/src/compose/rollups/docker-compose-espresso.yaml
```diff
@@ -16,6 +16,39 @@ services:
...
+        command: "cartesi-rollups-espresso-reader"
```
tuler:
```suggestion
        command: ["cartesi-rollups-espresso-reader"]
```

apps/cli/src/node/docker-compose-espresso.yaml
miltonjonat: Coming back to this from the discussions on #171, I'm now thinking that a path that would make sense would be http://localhost:8080/eip712/nonce and http://localhost:8080/eip712/submit 

What do you think?

tuler: EIP-712 is a typed structured data hashing and signing standard. I don't think it should be the name of the service.

tuler (approved):

PR #152, opened 2025-02-04
feat(cli): validator container runs with cartesi unprivileged user

endersonmaia:
Using the post_start docker compose hook, we can create the snapshot directory and copy the snapshot using the root user and start the container with cartesi unprivileged user.

Requires compose plugin 2.30.0

See: https://docs.docker.com/compose/how-tos/lifecycle/

---

This pull request includes several updates to the `@cartesi/cli` package, focusing on improving security and updating dependencies. The most important changes include running the validator container with an unprivileged user and updating the minimum required Docker Compose version.

Security improvements:

* [`apps/cli/src/node/docker-compose-validator.yaml`](diffhunk://#diff-97736c1f31d49c10ad0362dccd705aaf822d05b3ad0f70bb0f3770cbab50fb37L16-R30): Modified the `services` section to run the validator container with the `cartesi` unprivileged user instead of `root`.

Dependency updates:

* [`apps/cli/src/commands/doctor.ts`](diffhunk://#diff-689a19d361b510cae4456bc4f8c786bd5dd61176208db0eb9e8811ea377607c5L11-R11): Updated the minimum required Docker Compose version to `2.30.0`.

Documentation:

* [`.changeset/seven-seals-pull.md`](diffhunk://#diff-180ec24d21d0ca1fb2e9a9de315d90b9b2a9d138f02a5e31c873b9e05c6836e9R1-R5): Added a changeset entry to document the update to run the validator container with the `cartesi` unprivileged user.


apps/cli/src/node/docker-compose-validator.yaml
```diff
@@ -13,13 +13,21 @@ services:
...
+                      mkdir -p "$CARTESI_SNAPSHOT_DIR"
+                      cp --recursive /tmp/snapshot/* "$CARTESI_SNAPSHOT_DIR"
+              user: root
         command:
             - /bin/bash
             - -c
             - |
-                mkdir -p "$CARTESI_SNAPSHOT_DIR"
-                cp --recursive /tmp/snapshot/* "$CARTESI_SNAPSHOT_DIR"
+                while ! stat "$CARTESI_SNAPSHOT_DIR" &>/dev/null; do
```
tuler, line 28: This busy wait and coordination between the `command` and `post_start` is not very elegant.
Does it really need to be this way? What if the `post_start` fails for whatever reason?

tuler:
In any case, if we move forward with a system-wide `cartesi start` we will need to revisit how the node get access to the cartesi machine of a specific project.

apps/cli/src/node/docker-compose-validator.yaml
endersonmaia: We could add some timeout to ensure it fails.

> Does it really need to be this way?

It's my approach to run the validator with an unprivileged user leveraging compose.

There are other approaches, for sure, would have to test some alternatives.

tuler: How about we first see how the `start` + `deploy` would work. Because it will be different in that case.

endersonmaia: sure, we can put this on-hold

if the rollups-node container image release already has this in-place, this process or preparing a directory should just be unnecessary

tuler:
Is there still a reason for this PR?

endersonmaia:
> Is there still a reason for this PR?

No, since rollups-node doesn't have a container image release anymore. 

Se could aim to run all services with unorivileged users, on another PR.

PR #177, opened 2025-03-13
Improve time to start

endersonmaia:
**TL;DR**

With first 2 items, already reduced from ~30s to ~9.5s on my environment.

- [X] Removed prompt service
- [X] Adjusted healthcheck intervals
- [x] Preload database (depends on #178)
- [ ] Move messages from removed prompt elsewhere (maybe `cartesi status` and on another PR?) 

---

This pull request includes significant changes to the `apps/cli` directory, focusing on removing the `prompt` service from various Docker Compose files and adjusting health check configurations.

### Removal of `prompt` service:

* [`apps/cli/src/commands/rollups/start.ts`](diffhunk://#diff-c73013ce1f29fd457c9da8213e3749c98860f0730d5bbe2978261d77347fa30dL97): Removed the `docker-compose-prompt.yaml` reference from the list of Docker Compose files.
* [`apps/cli/src/compose/rollups/docker-compose-anvil.yaml`](diffhunk://#diff-97e1407e72d5923e26ee3c52733166478fc3bca39f2dcc8fc781937cb4ad03b4L13-L24): Removed the `prompt` service configuration.
* [`apps/cli/src/compose/rollups/docker-compose-bundler.yaml`](diffhunk://#diff-896f5f48466a0051cd49bd3b49b9b385bce9aa7af7bdadae129e98dc5fdbd415L33-L36): Removed the `prompt` service configuration.
* [`apps/cli/src/compose/rollups/docker-compose-database.yaml`](diffhunk://#diff-ab80ca76fe4a2efe23ed8c048ebc142d1f99d8c9fb5b4995e835a6404db64bdfL6-R8): Removed the `prompt` service configuration.
* [`apps/cli/src/compose/rollups/docker-compose-espresso.yaml`](diffhunk://#diff-0343d89d288e5a884ea7f3274623db9d8f05a9f75a5b12ebdaaa484e6f5cf8dfL50-L54): Removed the `prompt` service configuration.
* [`apps/cli/src/compose/rollups/docker-compose-explorer.yaml`](diffhunk://#diff-efe73958c750db050f2d879840c2133120965ed708ed8201c2226d581aded79aL53-L57): Removed the `prompt` service configuration.
* [`apps/cli/src/compose/rollups/docker-compose-graphql.yaml`](diffhunk://#diff-930f4b0b8ee548f6105ea28bce2fd9a5036f2f07b6488bc21e02f137cc401a89L40-L44): Removed the `prompt` service configuration.
[...]

PR #153, opened 2025-02-04
refactor(cli): better prompt dependencies

endersonmaia:
This pull request includes changes to improve the dependency management of the `prompt` service in various Docker Compose configurations. The most important changes are the addition of `depends_on` conditions to ensure that the `prompt` service starts only after its dependencies are in the appropriate state.

Dependency management improvements:

* [`.changeset/lazy-feet-explode.md`](diffhunk://#diff-81aafa642b25cb7c852567754ad647b8076966759e9032c36f10fe572c1777fdR1-R5): Added a changeset file indicating a patch update for `@cartesi/cli` with a note on better prompt service dependency.
* [`apps/cli/src/node/docker-compose-anvil.yaml`](diffhunk://#diff-4e2567b6365f4839be81ffba67c0960045900a89ef839a76ba08c8d7cb454719L38-R48): Added `depends_on` conditions for `anvil` and `dapp_deployer` services to ensure the `prompt` service starts only after these services are healthy or completed successfully.
* [`apps/cli/src/node/docker-compose-bundler.yaml`](diffhunk://#diff-20fd65c916a75b33482fdc2ab4f13908d68eed03378286fb8f771ee8c5dd9e7bL34-R36): Added `depends_on` condition for the `alto` service to ensure the `prompt` service starts only after the `alto` service has started.
* [`apps/cli/src/node/docker-compose-explorer.yaml`](diffhunk://#diff-5c92ef11e6978d2739845cbcb3d7fc67f75580d2a59bab6741321392689ed4b1L54-R56): Added `depends_on` condition for the `explorer` service to ensure the `prompt` service starts only after the `explorer` service has started.
* [`apps/cli/src/node/docker-compose-paymaster.yaml`](diffhunk://#diff-dd9ab5bc6dedd4e561f4bd0b327aebc454e0df48ed1e552bdb6e9c6cee02c52cL10-R12): Added `depends_on` condition for the `mock-verifying-paymaster` service to ensure the `prompt` service starts only after the `mock-verifying-paymaster` service has started.
* [`apps/cli/src/node/docker-compose-validator.yaml`](diffhunk://#diff-97736c1f31d49c10ad0362dccd705aaf822d05b3ad0f70bb0f3770cbab50fb37L31-R33): Added `depends_on` condition for the `validator` service to ensure the `prompt` service starts only after the `validator` service is healthy.

PR #179, opened 2025-03-14
Replacement of prompt service with custom code

tuler:
While working on the startup feedback, I ended up redoing a lot of the work that was done at PRs https://github.com/cartesi/cli/pull/177 and https://github.com/cartesi/cli/pull/153
So this PR includes some of those changes, but also the new startup feedback.
Please check my individual commits on this PR, and compare them to the ones at https://github.com/cartesi/cli/pull/177

The startup feedback works by creating a [listr2](https://listr2.kilic.dev) of tasks where each task is a polling of `docker compose ps <service>` for the healthy status of a service. For each service there is a service name, but also a `healthySemaphore` which is the name of the service that indicates healthy status.

For example, the `graphql` is a service, but we need to wait for the `proxy` service to be healthy to report access to `graphql`, as it is proxied through it.

I only implemented healthy status checks, which means that every service that is part of the startup feedback will now require a health check configuration.  We are still missing health check config for the following services:

- [x] health check for espresso
- [x] health check for explorer
- [x] health check for bundler
- [x] health check for paymaster


tuler:
Total startup time in my machine with `--services graphql` is taking from 5 to 6 seconds.

apps/cli/src/compose/rollups/docker-compose-graphql.yaml
```diff
@@ -33,10 +33,20 @@ services:
         expose:
             - 8080
         command: ["cartesi-rollups-graphql"]
+        healthcheck:
+            test: ["CMD", "curl", "-f", "http://127.0.0.1:8080/health"]
+            start_period: 10s
+            start_interval: 200ms
+            interval: 10s
+            timeout: 5s
```
endersonmaia: Should we use similar values for all other services?

Default `interval` is `30s`, I'm ok with that.

Since everything is local, a `timeout` of `1s` is more then enough, the default value is also `30s`.

apps/cli/src/commands/rollups/start.ts
```diff
@@ -1,20 +1,133 @@
...
+    healthySemaphore?: string; // service to check if the service is healthy
+    healthyTitle?: string | ((port: number) => string); // title of the service when it is healthy
+    waitTitle?: string; // title of the service when it is starting
+    errorTitle?: string; // title of the service when it is not healthy
+};
+
+const host = "http://127.0.0.1";
+
+// services configuration
+const standardServices: Service[] = [
```
endersonmaia:
```suggestion
const baseServices: Service[] = [
```

I prefer base (since I already started using it on another PR :sweat_smile: ), but no hard opinions.

endersonmaia (changes requested):
Missing healthchecks for services you described and defined the dependency.

apps/cli/src/compose/rollups/docker-compose-graphql.yaml
tuler: I'm ok with the default, 30s

endersonmaia: > I'm ok with the default, 30s

For `timeout` or `interval` ?

tuler: For interval

endersonmaia (changes requested):
I suggest normalizing healthcheck parameters between services (interval, start_time, etc.)

tuler:
> I suggest normalizing healthcheck parameters between services (interval, start_time, etc.)

Done.
Is timeout `1a` good enough for `pg_isready`?

endersonmaia (approved):
`--verbose` is not working, but I'll consider this PR approved, and we can tackle this --verbose behavior on another PR

endersonmaia (approved):

PR #185, opened 2025-03-17
Change verbose behavior

tuler:
This removes the `rollups start --verbose` flag.
Log level of rollups-node is set to "info".

Maybe we should still have the flag, and set rollups-node level to "debug" in case of verbose?

### DIFF
--- a/.changeset/evil-colts-rule.md
+++ b/.changeset/evil-colts-rule.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": major
+---
+
+add rollups logs command
--- a/.changeset/kind-chicken-take.md
+++ b/.changeset/kind-chicken-take.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": major
+---
+
+change verbose behavior
--- a/apps/cli/src/commands/rollups.ts
+++ b/apps/cli/src/commands/rollups.ts
@@ -1,5 +1,6 @@
 import { Command } from "@commander-js/extra-typings";
 import { createDeployCommand } from "./rollups/deploy.js";
+import { createLogsCommand } from "./rollups/logs.js";
 import { createStartCommand } from "./rollups/start.js";
 import { createStatusCommand } from "./rollups/status.js";
 import { createStopCommand } from "./rollups/stop.js";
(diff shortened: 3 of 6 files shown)


endersonmaia:
I think that, since we now have a start command defaulting to detached mode from compose, we could have a cartesi rollups logs command instead

tuler:
I agree we can have a `cartesi rollups logs`.
But do we need a `--verbose` to turn on more verbose log from rollups-node?

tuler:
Added logs command to this same PR

endersonmaia:
> But do we need a --verbose to turn on more verbose log from rollups-node?

We eventually need this when we need to debug. We could think of an alternative way of debugging, outside the cli.

Maybe generating the `compose.yaml` with `--dry-run` and calling `docker compose up` after making changes to increase debug levels.

We could enable manipulating environment variables like we do with `.cartesi.env` and document that this will impact the `start` command.

tuler:
> We eventually need this when we need to debug. We could think of an alternative way of debugging, outside the cli.

How about this?
https://github.com/cartesi/cli/pull/185/commits/a06787654cda3cdaff9ed45b00036389dab19754

endersonmaia:
> > We eventually need this when we need to debug. We could think of an alternative way of debugging, outside the cli.
> 
> How about this? [a067876](https://github.com/cartesi/cli/commit/a06787654cda3cdaff9ed45b00036389dab19754)

If the idea is control only the rollups-node service, I think it's enough.

endersonmaia (approved):
If I got it right, we'll have `start` getting servicves up and showing their progress until thre're ready and their endpoints.

`--verbose` will impact the `logs` command output but not the `start` output.

In that case, LGTM.

endersonmaia (changes requested):
Change the commit message (and chageset) accordingly, since `--verbose` won't be removed.

tuler:
> If I got it right, we'll have `start` getting servicves up and showing their progress until thre're ready and their endpoints.
> 
> `--verbose` will impact the `logs` command output but not the `start` output.

No. The `verbose` will impact the log of the service and must be defined in the `start`.

endersonmaia (approved):

PR #197, opened 2025-03-19
ci: pin GitHub Actions commit hash

endersonmaia:
See: https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

Since we already have dependabot acting on github-actions, whenever some GH action has an update to the version, we shall get a PR to update it to the new commit hash.


endersonmaia:
Closed in favor of #198

PR #199, opened 2025-03-20
Change espresso endpoint to /transaction

tuler:
Tested with:

```
curl -X POST http://127.0.0.1:8080/transaction/nonce -H "Content-Type: application/json" -d '{"msg_sender":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","app_contract":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"}'
```

Did not test with `/transaction/submit`, but we can't use help to do that after we merge and release.

Also removed the exposure of espresso-dev-node to host through proxy.


endersonmaia (approved):
If there's no use for the espresso ports to be exposed to localhost, we could also remove them from compose.

endersonmaia (approved):
Don't forget to squash.

PR #204, opened 2025-03-21
Add cartesi/rollups-database container image release

endersonmaia:


tuler (commented):
Do we need a new workflow for each image, or is it better to have a single workflow for the "sdk" project that takes care of all images, because they follow the same trigger and same versioning.

endersonmaia:
> Do we need a new workflow for each image, or is it better to have a single workflow for the "sdk" project that takes care of all images, because they follow the same trigger and same versioning.

Yeah, I'll send a fixup and see how it goes.

endersonmaia:
> Do we need a new workflow for each image, or is it better to have a single workflow for the "sdk" project that takes care of all images, because they follow the same trigger and same versioning.

I couldn't manage to make `docker/metadata-action` define the tag and annotations for multiple bake targets without making it more complicated than just replicating the workflow files.

tuler (approved):

PR #205, opened 2025-03-21
cli to use pre-initialized database

endersonmaia:
When we have the container image releases  from #209 and #204 we can change this PR to adapt to the names defined.

We could even use the `runtime` image to run the rollups-node service in the compose if we want to.


tuler:
Need to rebase to `prerelease/v2-alpha`

tuler (approved):

tuler (approved):

tuler (approved):

PR #206, opened 2025-03-24
ci(sdk): use ghcr.io registry cache

endersonmaia:
First run:

Screenshot from 2025-03-24 11-49-43

Second run:

Screenshot from 2025-03-24 11-53-24


tuler:
Opted for https://github.com/cartesi/cli/pull/208

PR #236, opened 2025-04-04
chore(sdk): move migration code to cartesi/rollups-runtime

endersonmaia:
This pull request includes several changes to the Cartesi SDK, primarily focusing on migration management and Dockerfile adjustments. The most important changes include the introduction of separate migration tables for the espresso-reader, the relocation of migrations to rollups-runtime, and updates to the Dockerfile to reflect these changes.

The objective is that people can use the `cartesi/rollups-runtime` image to run a node and use the migration script on an existing database or when `cartesi/rollups-database` won't be used.

Migration management improvements:

* [`.changeset/great-pens-report.md`](diffhunk://#diff-b205e52968d64ed55d15389931805518ed4a4020e543e3d483e0ed09bcd105a2R1-R5): Introduced a separate table to control espresso-reader migrations to avoid conflicts with rollups-node migrations.
* [`.changeset/small-coats-allow.md`](diffhunk://#diff-04d622dc80fa931d80a7a5cc58636c7d8880de1b4048665152b3c6cc13e3e64bR1-R5): Moved migration processes to rollups-runtime.

Dockerfile updates:

* [`packages/sdk/Dockerfile`](diffhunk://#diff-fbbb6309bd1c5bde85dc9baab5c5c1ef5af52fedb5e3fd28cecf37916d3ef1feR181-R183): Added new COPY commands to include migration binaries and scripts for the espresso-reader and graphql migrations.
* [`packages/sdk/Dockerfile`](diffhunk://#diff-fbbb6309bd1c5bde85dc9baab5c5c1ef5af52fedb5e3fd28cecf37916d3ef1feL231-R234): Updated the espresso-reader migration script to use the new migration table.
* [`packages/sdk/Dockerfile`](diffhunk://#diff-fbbb6309bd1c5bde85dc9baab5c5c1ef5af52fedb5e3fd28cecf37916d3ef1feL343-L345): Removed redundant COPY commands that were previously added.

### DIFF
--- a/.changeset/great-pens-report.md
+++ b/.changeset/great-pens-report.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/sdk": patch
+---
+
+use differente table to control espresso-reader migrations to avoid conflict with rollups-node migrations
--- a/.changeset/small-coats-allow.md
+++ b/.changeset/small-coats-allow.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/sdk": patch
+---
+
+move migration to rollups-runtime
--- a/packages/sdk/Dockerfile
+++ b/packages/sdk/Dockerfile
@@ -178,6 +178,9 @@ EOF
 
 COPY --from=graphql /usr/local/bin/cartesi-rollups-graphql /usr/local/bin/
 COPY --from=espresso-reader /usr/local/bin/cartesi-rollups-espresso-reader /usr/local/bin/
+COPY --from=go-migrate /usr/local/bin/migrate /usr/local/bin/
+COPY --from=graphql-migration /usr/share/cartesi/rollups-graphql/migrations /usr/share/cartesi/rollups-graphql/migrations
+COPY --from=espresso-reader-migration /usr/share/cartesi/rollups-espresso-reader/migrations /usr/share/cartesi/rollups-espresso-reader/migrations
 
 USER cartesi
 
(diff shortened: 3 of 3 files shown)


tuler:
What is the expected use of migrations in a production environment?
I was expecting that in a production environment migrations would be handled manually, not on startup of services.

endersonmaia:
> What is the expected use of migrations in a production environment? I was expecting that in a production environment migrations would be handled manually, not on startup of services.

Automate the process of bootstraping a news instance of a node.

I would only need to provide proper database credentials, but database creation and migration would be handled during initialization of the services.

tuler:
It's a bit unsafe that way. Ideally services that only read from the database, currently both rpc and graphql[1], would only need read credentials. Not if database setup is executed by the service.

[1] I'm actually not sure how is the current architecture, but it definitely only reads from the node database.

endersonmaia:
> It's a bit unsafe that way. Ideally services that only read from the database, currently both rpc and graphql[1], would only need read credentials. Not if database setup is executed by the service.
> 
> [1] I'm actually not sure how is the current architecture, but it definitely only reads from the node database.

rollups-graphl only needs to READ from the rollups-node database, but it needs to create it's own database do save the data that will be used to serve the graphql endpoint

espresso-reader needs access to create a least a schema in the same database as rollups-node

I can control which database credentials to provide to each service before it starts, but I´d like it to stop there.

Every other step that can be done by the application or its migration code, should be handled by them after proper credentials are given.

tuler:
changeset

tuler (approved):

PR #245, opened 2025-04-11
coprocessor create command

### LINKED ISSUE #227: Implement `cartesi coprocessor create`
The creation of a coprocessor project has some aditional steps comparing to a rollups.

That being said, we may need tro have separate create commands like:

- `cartesi rollups create`
- `cartesi coprocessor create`

**Steps**

- [ ] Deprecate the `cartesi create` command
- [ ] Move current `cartesi create` to `cartesi rollups create`
- [ ] Add `cartesi coprocessor create`

**Goal**

The goal of `cartesi coprocessor create` involves:

- creating a smart contracts directory from forge template
- copy CoprocessorAdapter
- adding the smartcontract for a coprocesssor application

My suggestion is to put all of this inside `cartesi/application-templates` so that we can replicate the logic for currente cartei create where we just download the diretory from the application-template repository for a specific `--template`

We should discuss if we need to have a specific branch or reorganize cartesi/application-tempalte repository to acoomodate rollups and coprocessor templates in the same repository with a hierachy differente from what we have now.

Reference: https://github.com/Mugen-Builders/co-processor-cli/blob/main/src/commands/create.rs

/cc @tuler

Nonnyjoe:
This PR closes #227

### DIFF
--- a/apps/cli/src/commands/coprocessor/create.ts
+++ b/apps/cli/src/commands/coprocessor/create.ts
@@ -29,14 +29,14 @@ export const createCreateCommand = () => {
                     "coprocessor",
                     template,
                     branch,
-                    `${name}/coprocessor_program`,
+                    path.join(name, "app"),
                 );
 
                 await download(
                     "coprocessor",
                     "solidity",
                     branch,
-                    `${name}/contracts`,
+                    path.join(name, "contracts"),
                 );
 
                 spinner.succeed(
--- a/apps/cli/src/commands/coprocessor/create.ts
+++ b/apps/cli/src/commands/coprocessor/create.ts
@@ -1,6 +1,7 @@
 import { Command, Option } from "@commander-js/extra-typings";
 import chalk from "chalk";
 import ora from "ora";
+import path from "path";
 import {
     DEFAULT_TEMPLATES_BRANCH,
     download,
@@ -10,6 +11,7 @@ import {
 export const createCreateCommand = () => {
     return new Command("create")
         .argument("<name>", "application and directory name")
+        .description("Create a coprocessor application template.")
         .addOption(
             new Option("-t, --template <template>", "template name to use")
                 .choices(TEMPLATES)
@@ -23,13 +25,23 @@ export const createCreateCommand = () => {
         .action(async (name, { branch, template }) => {
             const spinner = ora("Creating application...").start();
             try {
-                const { dir } = await download(
-                    "rollups",
+                let { dir } = await download(
+                    "coprocessor",
                     template,
                     branch,
-                    name,
+                    `${name}/coprocessor_program`,
                 );
-                spinner.succeed(`Application created at ${chalk.cyan(dir)}`);
+
+                await download(
+                    "coprocessor",
+                    "solidity",
+                    branch,
+                    `${name}/contracts`,
+                );
+
+                spinner.succeed(
+                    `Application created at ${chalk.cyan(path.dirname(dir))}`,
+                );
             } catch (e: unknown) {
                 spinner.fail(
                     e instanceof Error
--- a/apps/cli/src/template.ts
+++ b/apps/cli/src/template.ts
@@ -22,3 +22,17 @@ export const download = async (
         providers: { cartesi: cartesiProvider },
     });
 };
+
+export const DEFAULT_TEMPLATES_BRANCH = "prerelease/sdk-12";
+
+export const TEMPLATES = [
+    "cpp",
+    "cpp-low-level",
+    "go",
+    "javascript",
+    "lua",
+    "python",
+    "ruby",
+    "rust",
+    "typescript",
+];
--- a/.changeset/five-keys-clap.md
+++ b/.changeset/five-keys-clap.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": major
+---
+
+NEW: coprocessor create command
--- a/apps/cli/src/commands/coprocessor.ts
+++ b/apps/cli/src/commands/coprocessor.ts
@@ -1,4 +1,5 @@
 import { Command } from "@commander-js/extra-typings";
+import { createCreateCommand } from "./coprocessor/create.js";
 
 export const createCoprocessorCommand = () => {
     const command = new Command("coprocessor")
@@ -10,6 +11,7 @@ export const createCoprocessorCommand = () => {
         .action(async (_options, program) => {
             program.help();
         });
+    command.addCommand(createCreateCommand());
     return command;
 };
 


tuler:
rebase on top of [feature/coprocessor-cli](https://github.com/cartesi/cli/tree/feature/coprocessor-cli)

endersonmaia:
rebased

/cc @Nonnyjoe

apps/cli/src/commands/coprocessor/create.ts
```diff
@@ -0,0 +1,70 @@
...
+    "javascript",
+    "lua",
+    "python",
+    "ruby",
+    "rust",
+    "typescript",
+] as const;
+
+const subDirs = ["coprocessor_program", "contracts"];
+let route: string;
```
endersonmaia:
```suggestion
const subDirs = ["coprocessor_program", "contracts"];
let route: string;
```

apps/cli/src/commands/coprocessor/create.ts
```diff
@@ -0,0 +1,70 @@
...
+                            "solidity",
+                            branch,
+                            `${name}/${subdir}`,
+                        );
+                    }
+                }
+                spinner.succeed(
+                    `Application created at ${chalk.cyan(path.dirname(route))}`,
+                );
+            } catch (e: unknown) {
```
endersonmaia: **UNTESTED CODE**

```suggestion
            try {

                let { dir } = await download(
                    "coprocessor",
                    template,
                    branch,
                    `${name}/coprocessor_program`,
                );
                dir = dir.replace(/\/src$/, "");

                await download(
                    "coprocessor",
                    "solidity",
                    branch,
                    `${name}/contracts`,
                );

                spinner.succeed(
                    `Application created at ${chalk.cyan(dir)}`,
                );
            } catch (e: unknown) {
```

A simplyfied suggestion.
```suggestion
            try {
                for (const subdir of subDirs) {
                    if (subdir === "coprocessor_program") {
                        const { dir } = await download(
                            "coprocessor",
                            template,
                            branch,
                            `${name}/${subdir}`,
                        );
                        route = dir;
                    } else {
                        const { dir } = await download(
                            "coprocessor",
                            "solidity",
                            branch,
                            `${name}/${subdir}`,
                        );
                    }
                }
                spinner.succeed(
                    `Application created at ${chalk.cyan(path.dirname(route))}`,
                );
            } catch (e: unknown) {
```

I'm not happy with my `dir = dir.replace(/\/src$/, "");`, but that's it.

endersonmaia (changes requested):

apps/cli/src/commands/coprocessor/create.ts
```diff
@@ -0,0 +1,70 @@
...
+    "cpp",
+    "cpp-low-level",
+    "go",
+    "javascript",
+    "lua",
+    "python",
+    "ruby",
+    "rust",
+    "typescript",
+] as const;
```
endersonmaia: I think this definitions could be centralized at a common `apps/cli/commands/create.ts`, or any other better place.

endersonmaia (changes requested):

apps/cli/src/commands/coprocessor/create.ts
endersonmaia: there's `app/cli/src/templates.ts`, that could be the better place

tuler:
Have we considered merging the application code and solidity code in a single directory?

A foundry project is just a few files. The committed files are only:
- foundry.toml
- contracts directory (instead of src)
- soldeer.lock
And a few git ignored output files/directories.

This would make the developer life easier, as he can run `cartesi build` and `forge build` from the same directory.
Smart contracts will be separate in a `contracts` directory.

Nonnyjoe:
Yeahhh @tuler I've considered that previously but the problem with that is that it makes the repo appear bloated in large applications. 

Some applications would need extra solidity codes like custom NFT's, Vaults etc, and having all this in the same directory as the Cartesi program that on its own has the potential to become very large might be problematic. Plus the fact that after building the contract you have an extra 'out' directory with the artifact of all the contracts and libraries. I just concluded that it looks cleaner and better to manage if we have these in different folders.

Plus most teams would prefer to have the frontend, solidity code and coprocessor program in separate folders in the same repo especially since it's easier for different team members to handle.

tuler:
> Some applications would need extra solidity codes like custom NFT's, Vaults etc, and having all this in the same directory as the Cartesi program that on its own has the potential to become very large might be problematic

But they are all inside the `contracts` directory, not "bloating" everything else.

Nonnyjoe:
> > Some applications would need extra solidity codes like custom NFT's, Vaults etc, and having all this in the same directory as the Cartesi program that on its own has the potential to become very large might be problematic
> 
> But they are all inside the `contracts` directory, not "bloating" everything else.

Yeah.... It'll definitely look bloated when you start considering: 
- dependency folder, 
- out folder, 
- scripts folder, 
- tests folder, 
- cache folder, 
- remappings.txt 

Some of these are not visible now but would appear once you build your contract, so having all this mix with other cargo files and folder for a rust application will definitely not look nice.

apps/cli/src/commands/coprocessor/create.ts
Nonnyjoe: Yeahhh, for the `dir = ....` we could use `path.dirname(dir);` to output the parent directory,

apps/cli/src/commands/coprocessor/create.ts
```diff
@@ -0,0 +1,51 @@
+import { Command, Option } from "@commander-js/extra-typings";
+import chalk from "chalk";
+import ora from "ora";
+import path from "path";
+import { download, templates } from "../../template.js";
+
+export const DEFAULT_TEMPLATES_BRANCH = "prerelease/sdk-12";
```
endersonmaia: this can also me moved to `apps/cli/src/template.ts`

apps/cli/src/template.ts
```diff
@@ -22,3 +22,17 @@ export const download = async (
...
+        "cpp-low-level",
+        "go",
+        "javascript",
+        "lua",
+        "python",
+        "ruby",
+        "rust",
+        "typescript",
+    ];
+};
```
endersonmaia: let's keep constants upper case, and it doesn't need to be a function, remember to change this also on `apps/cli/src/commands/rollups/create.ts` for consistency.

```suggestion
export const TEMPLATES = [
        "cpp",
        "cpp-low-level",
        "go",
        "javascript",
        "lua",
        "python",
        "ruby",
        "rust",
        "typescript",
    ];
```

endersonmaia (changes requested):

endersonmaia (approved):
LGTM

endersonmaia (changes requested):
I miss the changeset file that was added when I rebased. (sorry about that)

endersonmaia (approved):
LGTM

You can squash both commit into a single one.

Nonnyjoe:
> You can squash both commit into a single one.

Done

apps/cli/src/commands/coprocessor/create.ts
```diff
@@ -0,0 +1,53 @@
...
+            DEFAULT_TEMPLATES_BRANCH,
+        )
+        .action(async (name, { branch, template }) => {
+            const spinner = ora("Creating application...").start();
+            try {
+                let { dir } = await download(
+                    "coprocessor",
+                    template,
+                    branch,
+                    `${name}/coprocessor_program`,
```
tuler: use `path.join(name, "coprocessor_program")` instead

apps/cli/src/commands/coprocessor/create.ts
```diff
@@ -0,0 +1,53 @@
...
+                );
+
+                await download(
+                    "coprocessor",
+                    "solidity",
+                    branch,
+                    `${name}/contracts`,
```
tuler: use `path.join(name, "contracts")` instead

apps/cli/src/commands/coprocessor/create.ts
tuler: Is there a better name for this directory?
Maybe just "app"?

apps/cli/src/commands/coprocessor/create.ts
```diff
@@ -0,0 +1,53 @@
...
+                );
+
+                spinner.succeed(
+                    `Application created at ${chalk.cyan(path.dirname(dir))}`,
```
tuler, line 43: Do you mean `dir` or `name`? That can confuse the user.

Nonnyjoe: I meant 'dir' actually, the log is suppose to return the path to the newly created project.
It should look like this: 
``` 
✔ Application created at /Users/idogwuchi/Documents/Cartesi/test-update
```
Where 'test-update' is the name of the application

apps/cli/src/commands/coprocessor/create.ts
Nonnyjoe: > Is there a better name for this directory? Maybe just "app"?

Sure we could use "app", I'll work on it.

apps/cli/src/commands/coprocessor/create.ts
Nonnyjoe: > ```
> ✔ Application created at /Users/idogwuchi/Documents/Cartesi/test-update
> ```

Would you like me to change this @tuler ?

tuler: No, that's fine.

tuler (approved):
LGTM.

One thing that may bite us in the future is the lack of README.md files in the generated application. Adding those to the `app` and `contracts` is easy, just a matter of adding them to `application-templates` repo. But if we need a README.md in the root directory we'll need to find another solution.

Nonnyjoe:
> One thing that may bite us in the future is the lack of README.md files in the generated application. Adding those to the `app` and `contracts` is easy, just a matter of adding them to `application-templates` repo. But if we need a README.md in the root directory we'll need to find another solution.

Yeahh.... That's true.
But would it be bad practice to have a util function that generates a README file, it can collect arguments like the app name and personalise the README file by adding the app name to it. A rough example could be:

```
function generateReadme(appName: string) {
    const readmeContent = `# ${appName}\n`;
    const outputPath = path.join(process.cwd(), "README.md");
    await fs.promises.writeFile(outputPath, readmeContent, "utf-8");
    .............
```

PR #250, opened 2025-04-17
coprocessor publish command

### LINKED ISSUE #230: Implement `cartesi coprocessor publish`
TBD

---

```
DRAFT

	- send the machine state to the solver
	- send car to ipfs (decide if we use ipfs public container or embed into cartesi/sdk)
	- check  :3034/ensure (SOLVER) if publish worked (healthcheck?)
```

Nonnyjoe:
This PR closes #230


tuler:
> * Added two more dependencies: (axios and form-data).

We try to minimize dependencies as much as possible. Can you use plain fetch instead?

Nonnyjoe:
> > * Added two more dependencies: (axios and form-data).
> 
> We try to minimize dependencies as much as possible. Can you use plain fetch instead?

Sure I can use plain fetch instead, but to avoid complexities, I think we should keep form-data

Nonnyjoe:
@tuler As requested I've removed both Axios and form-data.
Now making use of the native solution for both of them

tuler:
Coprocessor initiative on hold.

Issue #231, opened 2025-03-28
Implement `cartesi coprocessor deploy`


endersonmaia:
This command should deploy the application smartcontracts do the local devnet.

It's debatable if we should implement this or just guide the developer to foundry tools for that.

Nonnyjoe:
Hiii... @endersonmaia any final decision of if we'll be supporting the deploy command?
If yes, then I'll like to handle this also

Issue #156, opened 2025-02-17
Commands for Cartesi Coprocessor


endersonmaia:
We should provide an unified experience for Cartesi Rollups and Cartesi Coprocessor using the Cartesi CLI.

We can have a `rollups` command with a set of subcommands, and a `coprocessor` command with another set of subcommands.

The entire set of commands and subcommands would look like the following:

```
cartesi doctor
cartesi create
cartesi build
cartesi clean
cartesi hash

cartesi rollups address-book
cartesi rollups create -> cartesi create
cartesi rollups start
cartesi rollups deploy
cartesi rollups stop
cartesi rollups status
cartesi rollups send

cartesi coprocessor address-book
cartesi coprocessor create -> cartesi create
cartesi coprocessor start
cartesi coprocessor deploy
cartesi coprocessor stop
cartesi coprocessor status
cartesi coprocessor send
```

## doctor

Can work the same as it does now.

## create

Can have two params:
- framework: [rollups | coprocessor]
- language: cpp, python, javascript, etc

It can also be more interactive, asking questions if no params are specified.
The template can also have a `cartesi.toml` file that has the `framework` information, which can help with following commands in the dev workflow.

## build

Can work the same way with both frameworks if we follow the Dockerfile strategy. But maybe coprocessor has some extra steps.

## clean

Can work the same as it does now.

## hash

Can work the same as it does now.

## rollups address-book

Prints smart contracts addresses that support Cartesi Rollups.

## rollups create -> cartesi create

Create a Cartesi Rollups project. Can be an alias for `cartesi create --framework rollups`.

## rollups start

Start a Cartesi Rollups local environment.

## rollups deploy

Deploy the application to the Cartesi Rollups node.

## rollups stop

Stop the Cartesi Rollups local environment.

## rollups status

Check the status of the Cartesi Rollups local environment, including the deployed applications.

## rollups send

Sends an input to the Cartesi Rollups application.

## coprocessor address-book

Prints smart contracts addresses that support Cartesi Coprocessor.

## coprocessor create -> cartesi create

Create a Cartesi Coprocessor project. Can be an alias for `cartesi create --framework coprocessor`.

## coprocessor start

Start a Cartesi Coprocessor local environment.

## coprocessor deploy

Deploy the application to the Cartesi Coprocessor node.

## coprocessor stop

Stop the Cartesi Coprocessor local environment.

## coprocessor status

Check the status of the Cartesi Coprocessor local environment, including the deployed applications.

## coprocessor send

[...]

PR #251, opened 2025-04-24
coprocessor deploy command

Nonnyjoe:
This PR closes: #231 

It runs a combination of the Carfile generation, publication and finally solidity contract deployment.
The current code implementation is able to find the respective folders for the solidity or coprocessor files provided the command is called from either, the root folder, the contract folder, or the coprocessor program folder.

Deployment is done using Viem and would require the user to pass in the name of the contract and also optional constructor arguments


apps/cli/src/commands/coprocessor.ts
```diff
@@ -1,4 +1,6 @@
 import { Command } from "@commander-js/extra-typings";
+import { createDeployCommand } from "./coprocessor/deploy.js";
+import { createStartCommand } from "./coprocessor/start.js";
```
tuler: we removed start subcommand for now

tuler (changes requested):

apps/cli/src/compose/coprocessor/docker-compose-carize.yaml
tuler, line 1: we use compose to execute service of the start command.
if you just need a one shot execution you could just run the docker container, no need for a compose

apps/cli/src/compose/coprocessor/docker-compose-carize.yaml
```diff
@@ -0,0 +1,7 @@
+services:
+    carize:
+        image: ghcr.io/zippiehq/cartesi-carize:latest
```
tuler: we always use tagged versions, for stability

tuler: Can the car generation be implemented in typescript and remove the need for an outside docker dependency?
I played a lot with CAR in typescript before, and it's usually straightforward.
The tradeoff is adding another npm dependency vs downloading another docker dependency.

https://www.npmjs.com/package/ipfs-car

Nonnyjoe: Okay.... I'll check this out

apps/cli/src/commands/coprocessor/deploy.ts
```diff
@@ -0,0 +1,490 @@
...
+                    } else {
+                        return;
+                    }
+                }
+            } catch (e: unknown) {}
+        });
+};
+
+const publishCarFiles = async (cartesiDir: string): Promise<boolean> => {
+    const DEFAULT_OPERATOR_URL = "http://127.0.0.1:5001";
```
tuler: don't expect to have these ports exposed as the default.
they will be behind the `coprocessor start` proxy.

the method should receive an `options` object with these values.

apps/cli/src/commands/coprocessor/deploy.ts
```diff
@@ -0,0 +1,490 @@
...
+    const DEFAULT_SOLVER_URL = "http://127.0.0.1:3034";
+    const carFileName: string = "output.car";
+    const artifactsDir = path.resolve(cartesiDir, ".cartesi", "artifacts");
```
tuler: we avoid having `.cartesi` spread all over the code. That's why we created the `getContextPath` util method.

apps/cli/src/commands/coprocessor/deploy.ts
```diff
@@ -0,0 +1,490 @@
...
+                    }
+                }
+            } catch (e: unknown) {}
+        });
+};
+
+const publishCarFiles = async (cartesiDir: string): Promise<boolean> => {
+    const DEFAULT_OPERATOR_URL = "http://127.0.0.1:5001";
+    const DEFAULT_SOLVER_URL = "http://127.0.0.1:3034";
+    const carFileName: string = "output.car";
```
tuler: can we call this `image.car`?

apps/cli/src/commands/coprocessor/deploy.ts
```diff
@@ -0,0 +1,490 @@
...
+    const artifactsDir = path.resolve(cartesiDir, ".cartesi", "artifacts");
+    const carFilePath = path.resolve(artifactsDir, carFileName);
+    const cidFilePath = path.resolve(artifactsDir, "output.cid");
```
tuler: who defines the spec of the CAR?
Are these names `output.cid`, `output.size` expected by some other component?

tuler:
I think deploying with viem that way you are doing is not good.
We are getting in the way of the developer with something that maybe we should not.
See my [latest comment about that](https://github.com/cartesi/cli/issues/156#issuecomment-2828603199).

apps/cli/src/commands/coprocessor/deploy.ts
```diff
@@ -0,0 +1,490 @@
...
+                    options.constructorArgs == true ||
+                    options.constructorArgs == undefined
+                ) {
+                    constructorArgs = [];
+                } else {
+                    constructorArgs = options.constructorArgs;
+                }
+
+                let { cartesiDir, foundryDir } = await findSubdirectories();
+                if (cartesiDir && foundryDir) {
```
tuler, line 49: Code structure is too complex.
Should be something like:

```typescript
const car = await pack({ ... });
await publish({ car ... });
await deploy({ ... });
```

exceptions raised by any step will stop execution and show the error.

apps/cli/src/commands/coprocessor/deploy.ts
Nonnyjoe: Sure, it's possible, 
I'll work on that

apps/cli/src/commands/coprocessor/deploy.ts
Nonnyjoe: We can replace all reference of output with image.
I just need to confirm from @stskeeps that the operator is not requiring the file name to be output...

Nonnyjoe:
@tuler 
In line with your latest reviews, I've moved the Carize generation process to a Carize.ts file and this required the addition of 2 important dependencies, which I did add to the project.

I also implemented other reviews, like renaming the generated carfiles from "output..." to "image..."

apps/cli/src/carize.ts
```diff
@@ -0,0 +1,62 @@
+import chalk from "chalk";
+import { filesFromPaths } from "files-from-path";
+import fs from "fs-extra";
+import type { Block } from "ipfs-car";
+import { CAREncoderStream, createDirectoryEncoderStream } from "ipfs-car";
+import path from "path";
+import { Writable } from "stream";
+
+export const createCarFromDirectory = async (
```
tuler, line 17: This is an util function.
In an util function don't call `console.log` or do `process.exit`.
Return whatever you need to return, that is needed outside.

Nonnyjoe:
Had to rebase on top of the latest updates to `future/coprocessor-cli`, to avoid conflicts in the coprocessor.ts file and also to remove the previous commit that added the CI for carize generation

Nonnyjoe:
@tuler I made some updates to the Carize.ts in line with your reviews and also to make it fully compatible with the carfile unpacking process being used in the operator.

Nonnyjoe:
To test these modifications, you need to start the coprocessor devnet using the collowing steps:

- Clone the coprocessor repo with this command;
``` git clone  https://github.com/zippiehq/cartesi-coprocessor ```

- edit https://github.com/zippiehq/cartesi-coprocessor/blob/main/docker-compose-devnet.yaml#L29 this line to say:
```image: ghcr.io/zippiehq/cartesi-coprocessor-operator:pr-76```

- Run the commands:
``` 
docker compose -f docker-compose-devnet.yaml pull 

docker compose -f docker-compose-devnet.yaml up --wait -d
```

- Next you have to build this branch, create an alias, create a template, modify it, build the template and finally deploy your program using this a sample command similar to the below:

```
cartesi-dev coprocessor deploy MyContract --constructorArgs 0x8f86403A4DE0BB5791fa46B8e795C547942fE4Cf <machine hash>
```

**Note:** 
- Cartes-dev: Is an alias i used, you'll have to replace using yours.
- MyContract: is the name of the default contract in the template.
- "0x8f86403A4DE0BB5791fa46B8e795C547942fE4Cf" : Is the active task issuer address for the coprocessor ATM.

tuler:
Coprocessor initiative on hold.

PR #277, opened 2025-05-20
Install modern nodejs using nvm

tuler:
```
$ docker run cartesi/sdk:devel node --version
v22.15.1
$ docker run cartesi/sdk:devel npm --version
10.9.2
$ docker run cartesi/sdk:devel alto --version
0.0.4
$ docker run cartesi/sdk:devel passkey-server --help
Usage: passkey-server [options]

Options:
  -p, --port <port>  port to listen on (default: 3000)
  -h, --help         display help for command
```

### DIFF
--- a/packages/sdk/Dockerfile
+++ b/packages/sdk/Dockerfile
@@ -272,6 +272,8 @@ ARG CARTESI_MACHINE_EMULATOR_VERSION
 ARG CARTESI_PAYMASTER_VERSION
 ARG CARTESI_PASSKEY_SERVER_VERSION
 ARG CARTESI_ROLLUPS_NODE_VERSION
+ARG NODE_VERSION
+ARG NVM_VERSION
 ARG TARGETARCH
 ARG TARGETARCH
 ARG XGENEXT2_VERSION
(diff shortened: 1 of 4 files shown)


packages/sdk/Dockerfile
```diff
@@ -286,14 +287,23 @@ apt-get install -y --no-install-recommends \
...
     squashfs-tools \
     xxd \
     xz-utils
 rm -rf /var/lib/apt/lists/*
 EOF
 
+# Install nvm and node
+ENV NVM_DIR=/root/.nvm
+RUN <<EOF
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.3/install.sh | bash
```
endersonmaia: I don't like this piping a script from the internet to shell.

We could improve this by using the git commit hash in the url, at least we have reproducibility without having to trust nvm release process but still trusting GitHub.
```suggestion
curl --fail --proto '=https' --tlsv1.2 -o- https://raw.githubusercontent.com/nvm-sh/nvm/977563e97ddc66facf3a8e31c6cff01d236f09bd/install.sh | bash
```

The hash could be on a build arg if you will.


The ultimate process here would to git clone and verify the tag with gpg, but I won't complain if you don't.

endersonmaia (approved):
I made a comment about nvm installation and reproducibility and security, but won't block for it.

packages/sdk/Dockerfile
tuler: I don’t like piping to shell either, but I don’t like to deviate too much from the documented standard way of installing.

what is a good middle ground in this case?

endersonmaia: We need to run the `install.sh`, how we obtain and verify its integrity is up to us.

piping to shell is a convenient way, not the recommended way, I think.

endersonmaia: My suggestion was already a middle ground, by using the hash in the URL, we make sure there wasn't content modification, since tags can e changed.

And we trust GitHub is not messing around.

tuler: Every nvm installation instruction pipe to shell, that's what I meant by "recommended".
Updated with your suggestion.

packages/sdk/Dockerfile
```diff
@@ -286,14 +288,23 @@ apt-get install -y --no-install-recommends \
...
     squashfs-tools \
     xxd \
     xz-utils
 rm -rf /var/lib/apt/lists/*
 EOF
 
+# Install nvm and node
+ENV NVM_DIR=/root/.nvm
+RUN <<EOF
+curl -o- --fail --proto '=https' --tlsv1.2 -o- https://raw.githubusercontent.com/nvm-sh/nvm/${NVM_VERSION}/install.sh | bash
```
endersonmaia:
```suggestion
curl --fail --proto '=https' --tlsv1.2 -o- https://raw.githubusercontent.com/nvm-sh/nvm/${NVM_VERSION}/install.sh | bash
```

`-o-` is repeated

endersonmaia (changes requested):

endersonmaia (approved):

PR #278, opened 2025-05-20
Revert "fix(cli): keep evm-reader enabled"

endersonmaia:
This reverts commit 2949c5123663e009a97f24767fee35bc18100926.

---

My last understanding of this is that, when we enable `--services espresso` we should disable `CARTESI_FEATURE_INPUT_READER_ENABLED`.


tuler:
In theory we don't need to disable it.
Because the node is programmed to ignore espresso applications, and the espresso reader should be ignoring non-espresso applications.
We need to test this, and avoid disabling it.

endersonmaia:
espresso is not planned

PR #279, opened 2025-05-22
feat(cli): add hidden start --runtime-version option

endersonmaia:
Since it's internal use only option, I don't think it deserves a changeset.

This will allow testing a runtime built locally from  `packages/sdk`.

```
cartesi start --runtime-version devel --dry-run
```

### DIFF
--- a/apps/cli/src/commands/start.ts
+++ b/apps/cli/src/commands/start.ts
@@ -220,6 +220,7 @@ export const createStartCommand = () => {
                 port,
                 services,
                 verbose,
+                runtimeVersion,
             } = options;
             // path of the tool instalation
             const binPath = path.join(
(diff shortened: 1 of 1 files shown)


apps/cli/src/commands/start.ts
```diff
@@ -229,7 +237,7 @@ export const createStartCommand = () => {
                 CARTESI_ROLLUPS_NODE_CPUS: cpus?.toString(),
                 CARTESI_ROLLUPS_NODE_MEMORY: memory?.toString(),
                 CARTESI_SDK_IMAGE: DEFAULT_SDK_IMAGE,
-                CARTESI_SDK_VERSION: DEFAULT_SDK_VERSION,
+                CARTESI_SDK_VERSION: options.runtimeVersion.toString(),
```
tuler: Put the runtime version variable up there in line 220

apps/cli/src/commands/start.ts
```diff
@@ -237,7 +238,7 @@ export const createStartCommand = () => {
                 CARTESI_ROLLUPS_NODE_CPUS: cpus?.toString(),
                 CARTESI_ROLLUPS_NODE_MEMORY: memory?.toString(),
                 CARTESI_SDK_IMAGE: DEFAULT_SDK_IMAGE,
-                CARTESI_SDK_VERSION: options.runtimeVersion.toString(),
+                CARTESI_SDK_VERSION: runtimeVersion?.toString(),
```
tuler: Do you still need the `?.toString()`. Value should be `string`, not a `string | undefined`, as there is a defined default value.

tuler (approved):

PR #281, opened 2025-05-22
Optimize espresso-dev-node startup

### LINKED ISSUE #276: Long espresso-dev-node startup time
While testing this the espresso service takes a lot of time to be healthy.
The `start` wait is over after 1 minute, as expected, and espresso service is still trying to be up.
The `espresso` service is healthy after 2:40 minutes.
And still there are a LOT of errors in the console.
If all this is expected we can merge as is, especially because this is an optional service and experimental.
I don't think it's ready for non-alpha release though.

_Originally posted by @tuler in https://github.com/cartesi/cli/pull/263#pullrequestreview-2851237343_

tuler:
This PR tries to speed up espresso-dev-node startup time by using a new option called `ESPRESSO_DEV_NODE_L1_DEPLOYMENT=skip` only currently available at the `main` unreleased version of `espresso-dev-node`. It expects the smart contract to already be deployed to the L1 provider (anvil).

For that to work we first created a cannon package for the espresso network smart contracts.
That is available at https://github.com/cartesi/espresso-network/tree/cartesi
So the first step to test this PR is to clone that repo, switch to branch `cartesi` and run `cannon build` (assuming cannon is installed).

Then we integrated the cannon package into our devnet, which can be built using
```
pnpm run build --filter @cartesi/devnet
```
That assumes the cannon package `espresso-network:0.0.1@main` was built locally.
ps: when a stable version exists, we can publish to cannon registry.

Then we changed the binary used by the SDK to a `main` unreleased version.
Copy the unreleased devnet into SDK and build it.
```
cp -R packages/devnet/deployments packages/sdk
cp packages/devnet/anvil_state.json packages/sdk
pnpm run build --filter @cartesi/sdk
```

Then run the CLI with a `devel` image:
```
tsx apps/cli/src/index.ts start --services espresso --runtime-version devel
```


tuler:
CI will keep failing while cannon package is not published to registry

tuler:
My latest test showed this error (and a lot more after that).
```
ERROR hotshot_state_prover::service: Cannot sync the light client state, will retry: Error when communicating with the smart contract: error sending transaction: TransportError(ErrorResp(ErrorPayload { code: 3, message: "execution reverted: custom error 0x080ae8d9", data: Some(RawValue("0x080ae8d9")) }))
```

tuler:
Integrated into draft PR at https://github.com/cartesi/cli/pull/304

PR #295, opened 2025-06-11
Remove espresso support

endersonmaia:
Since there's still work to be done so that espresso-dev-node stability and contracts are ready to cartesi/cli, we should remove this from now, and wait for #276 to be done.


tuler:
We also have changes in the CLI.
We can prepare this PR, merge it, and then open a branch and revert the commit in that branch.

endersonmaia:
> We also have changes in the CLI.

Yes, I'm at it.

tuler (approved):
fixup, then merge

PR #298, opened 2025-06-12
Run command

tuler:
This `run` command will run a compose environment for a specific application (created with `cartesi create`), wait for the environment to be healthy, and immediately deploy the application if there is a build found in the project.

The name of the docker compose project will be given by docker compose itself, by following its own [naming rules](https://docs.docker.com/compose/how-tos/project-name/). It can also be explicitly defined using the `--project-name` run option. The name of the project will be used as the name registered with the `cartesi-rollups-node`. So client applications can use this name while interacting with the `cartesi-rollups-node`JSON-RPC API, instead of the application address (which will change on redeploy).

This version is exposing the cartesi machine snapshot in the host machine to the docker compose runtime environment by using [bind mounts](https://docs.docker.com/engine/storage/bind-mounts/), instead of copying files from the host to docker. This will avoid copying very large machines around in case of some applications.

If a build is not found on `run`, it just presents the warning:
```
machine snapshot not found, waiting for build
```

Then it shows a "shell" prompt, with 3 options.
```
(l) View logs   (b) Build and redeploy  (q) Quit
```

View logs shows the logs of the `cartesi-rollups-node` docker container.

Build and redeploy executes the `build` command, and then redeploy the application to the environment.
On every redeployment first we remove the previously deployed application, then a new deployment is performed.
The `salt` used is incremented, starting from 0, so even if the `hash` of the new machine is unchanged that will cause the deployment to a new address, of a new application, with a clean on-chain history.

The quit option will cleanup the environment, by shutting down all containers and deleting the volumes, leaving nothing behind. A CTRL-C should also shut down the environment gracefully.


endersonmaia:
Didn't review code yet, but running the changes.

When sending an input, I don't get the CLI to detect the application address anymore. I think send cloud accept the app name OR the address.

Also, when doing the build outside the `(b) Build and redeploy` console, what's the expected behavior?

The logs could be improved to know which app/templatehash was undeployed.

From:

`✔ my-go-dapp undeployed`

To:

`✔ my-go-dapp undeployed (address: 0x31d37ed4093c5e883844346e668fe28071698f00, machineHash: 359115a9664c57b3e2817ddc134d3b6aa74a4c7a4cacc623de108036990839d9)` 

Also, when running a second dapp, it fails with a port error from compose itself, it this error could be catch and reported nicely and not starting containers. Or includea message informing how to change the port already, or even just bumping the port number  like we discussed before. Whatever works best.

tuler:
> When sending an input, I don't get the CLI to detect the application address anymore. I think send cloud accept the app name OR the address.

One alternative is to send inputs by executing `docker compose exec rollups-node cartesi-rollups-cli send`.
That would use the save anvil as the node, instead of accessing anvil through RPC directly.
This is what I'm doing now with application deployment.

One problem is that that doesn't have support for portals.

tuler:
> The logs could be improved to know which app/templatehash was undeployed.
> 
> From:
> 
> `✔ my-go-dapp undeployed`
> 
> To:
> 
> `✔ my-go-dapp undeployed (address: 0x31d37ed4093c5e883844346e668fe28071698f00, machineHash: 359115a9664c57b3e2817ddc134d3b6aa74a4c7a4cacc623de108036990839d9)`

I wanted to avoid long lines in the console.
What happens is that the dapp is undeployed, whatever it is.

tuler:
> Also, when doing the build outside the `(b) Build and redeploy` console, what's the expected behavior?

I'm not monitoring the build anymore. That strategy was a little fragile IMO.
Maybe I can invest more time doing it in a another PR.

tuler:
> Also, when running a second dapp, it fails with a port error from compose itself, it this error could be catch and reported nicely and not starting containers. Or includea message informing how to change the port already, or even just bumping the port number like we discussed before. Whatever works best.

I'll introduce the port fix in another PR.

endersonmaia (approved):
I think `cartesi send` needs a fix to identify the app address or accept the app name, but won't block this PR for this, since one can just fill the app address and it works.

tuler:
Fixing outstanding issues in other PRs.

PR #301, opened 2025-06-16
Remove start and stop, fix send address

tuler:
Remove the `start` and `stop` commands, to focus only on `run`.
Also fix address resolution of `send`.


endersonmaia (approved):
LGTM

PR #304, opened 2025-06-16
Espresso support

tuler:
This PR adds support for espresso sequencer.

It tries to speed up espresso-dev-node startup time by using a new option called `ESPRESSO_DEV_NODE_L1_DEPLOYMENT=skip` only currently available at the `main` unreleased version of `espresso-dev-node`. It expects the smart contract to already be deployed to the L1 provider (anvil).

For that to work we first created a cannon package for the espresso network smart contracts.
That is available at https://github.com/cartesi/espresso-network/tree/cartesi
So the first step to test this PR is to clone that repo, switch to branch `cartesi` and run `cannon build` (assuming cannon is installed).

Then we integrated the cannon package into our devnet, which can be built using
```
pnpm run build --filter @cartesi/devnet
```
That assumes the cannon package `espresso-network:0.0.1@main` was built locally.
ps: when a stable version exists, we can publish to cannon registry.

Then we changed the binary used by the SDK to a `main` unreleased version.
Copy the unreleased devnet into SDK and build it.
```
cp -R packages/devnet/deployments packages/sdk
cp packages/devnet/anvil_state.json packages/sdk
pnpm run build --filter @cartesi/sdk
```

Then run the CLI with a `devel` image:
```
tsx apps/cli/src/index.ts start --services espresso --runtime-version devel
```


brunomenezes:
We are doing a grooming. I will be closing this one as `not planned`.

PR #306, opened 2025-06-18
Simplified send and deposit commands

tuler:
This PR brings changes to the `send` command and subcommands.

The following are the changed commands compared to CLI V1:

```diff
-cartesi send generic
+cartesi send

-cartesi send erc20
+cartesi deposit erc20

-cartesi send erc721
+cartesi deposit erc721

-cartesi send ether
+cartesi deposit ether

-cartesi send dapp-address
```

The inputs are sent directly to the anvil service. A viem `TestClient` is instantiated, instead of a `PublicClient` and `WalletClient` pair. This allows impersonation (sending inputs from any sender), which is good for testing.

The URL of anvil can be explicitly defined by the `--rpc-url` option, but is also automatically resolved if the command is executed inside a directory that has a running environment with `cartesi run`. This is implemented by using the `docker compose port proxy 8088`command, because the port can be defined by the user.

The deposit commands have been improved, and now balances are checked and approvals are automatically handled before transfer is performed.

As CLI V2 is focused on dev, ability to send inputs to other networks have been removed.


endersonmaia:
If I run on a folder that's not the one the dapp was executed with `cartesi run`, I get an error.

I think we could first verify if there's a compose project with that name (`docker compose ls` ? ) to fail clean.

```shell
❯ cartesi send
Command failed with exit code 1: docker compose --project-name cli port proxy 8088

service "proxy" is not running
```

endersonmaia:
Also, when trying to run `cartesi deposit` and I pick the erc721 it fails with:

```
❯ cartesi deposit
✔ Select type of asset to deposit erc721
error: missing required argument 'token-id'
```

But it never asked me what's this `token-id`.

endersonmaia (changes requested):
I made two comments:

- clean error when running send from outside the projects's directory
- check `send erc721`

tuler:
> * clean error when running send from outside the projects's directory

In that scenario I changed to ask for the RPC URL

tuler:
> * check `send erc721`

This is because the asset types subcommands are not consistent.
Check:

```
cartesi-dev deposit ether --help
cartesi-dev deposit erc20 --help
cartesi-dev deposit erc721 --help
```

Maybe the `amount` should be an argument for `ether` and `erc20`, as token-id is for `erc721`.

And also I could ask interactively if not defined instead of raising an error. So when it comes from `cartesi deposit` it wouldn't just exit with an error.

tuler:
> Maybe the `amount` should be an argument for `ether` and `erc20`, as token-id is for `erc721`.
> 
> And also I could ask interactively if not defined instead of raising an error. So when it comes from `cartesi deposit` it wouldn't just exit with an error.

Implemented this.

endersonmaia (approved):
squash before merge

PR #313, opened 2025-07-15
Chore: Bump rollups-explorer image

brunomenezes:
### Summary
Code changes in that image add support for adapting to port changes (e.g., 6751, 6752, and so on).

### DIFF
--- a/apps/cli/src/compose/docker-compose-explorer.yaml
+++ b/apps/cli/src/compose/docker-compose-explorer.yaml
@@ -78,7 +78,10 @@ services:
                 condition: service_healthy
 
     explorer:
-        image: cartesi/rollups-explorer:1.0.2
+        image: cartesi/rollups-explorer:1.2.0
+        environment:
+            NODE_RPC_URL: "http://127.0.0.1:${CARTESI_LISTEN_PORT:-6751}/anvil"
+            EXPLORER_API_URL: "http://127.0.0.1:${CARTESI_LISTEN_PORT:-6751}/explorer-api/graphql"
         expose:
             - 3000
         depends_on:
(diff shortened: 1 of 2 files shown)


apps/cli/src/compose/docker-compose-explorer.yaml
```diff
@@ -78,7 +78,10 @@ services:
                 condition: service_healthy
 
     explorer:
-        image: cartesi/rollups-explorer:1.0.2
+        image: cartesi/rollups-explorer:1.2.0
+        environment:
+            NODE_RPC_URL: "http://127.0.0.1:${CARTESI_LISTEN_PORT:-6751}/anvil"
+            EXPLORER_API_URL: "http://127.0.0.1:${CARTESI_LISTEN_PORT:-6751}/explorer-api/graphql"
```
endersonmaia, line 84: Why didnt you used explorer_api and anvil as hostnames an chose to go thrugh proxy?

endersonmaia (approved):
LGTM

apps/cli/src/compose/docker-compose-explorer.yaml
tuler: Because these endpoints are used by the browser when the explorer SPA is loaded.

tuler (approved):

apps/cli/src/compose/docker-compose-explorer.yaml
brunomenezes: @endersonmaia, Just complementing, that alias is used in the server-to-server communication [ref here](https://github.com/cartesi/rollups-explorer/blob/main/apps/web/.env.cartesi#L5)

PR #319, opened 2025-07-18
Additional config flags to drive machine build

tuler:
New config for `cartesi.toml`
- use_docker_env: inject docker image ENV into cartesi-machine ENV (default: true)
- use_docker_workdir: inject docker image WORKDIR into cartesi-machine WORKDIR (default: true)


endersonmaia (approved):
LGTM

PR #320, opened 2025-07-21
Possibility to use an override config file

tuler:
This PR allows the use of multiple config files for build, with an override strategy.
This enables something like:

```
cartesi-dev build -c cartesi.toml -c cartesi.mainnet.toml
```

Where `cartesi.toml` contains common configuration, and `cartesi.mainnet.toml` only contains values that should complement or override values in `cartesi.toml`. This helps to reduce code duplication, improving maintenance, like in [this commit](https://github.com/cartesi/honeypot/pull/26/commits/7ff3fe5352bea6b3dfe8d9075e5cb1065e619d8e).

This PR also removes 3 properties from the config that I think could lead to confusion: `interactive`, `store` and `finalHash`

### DIFF
--- a/.changeset/chubby-eagles-mate.md
+++ b/.changeset/chubby-eagles-mate.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+possibility to use an override config file
--- a/.changeset/curly-dolls-cross.md
+++ b/.changeset/curly-dolls-cross.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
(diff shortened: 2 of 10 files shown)


endersonmaia:
Using honeypot branch to test, I get the following error but it's not clear what is the problem:

```
❯ cartesi-dev build -c cartesi.toml -c cartesi.mainnet.toml
✔ Build drives
  ✔ Build drive root
  ✔ Build drive state
✖ Invalid state: WritableStream is locked
Invalid state: WritableStream is locked
```

endersonmaia (approved):
LGTM

PR #323, opened 2025-07-22
Fix ram_image default value

tuler:
This PR let the default value of the `ram-image` to be driven by the `cartesi-machine` CLI tool (used by `build`).
If `cartesi-machine` is installed in the system (macOS or Linux) the default location of the `linux.bin` will be defined by the `cartesi-machine` installation.
If the CLI uses the `cartesi-machine` from the SDK, then it uses the default value of that.
Either way, if the `cartesi.toml` does not define the value of `ram_image` it will work regardless we use the system `cartesi-machine` or the docker sdk `cartesi-machine`.

Note that if the user defines a `ram_image` in the `cartesi.toml` that path will be used. If it's a system `cartesi-machine` the path will be considered a host path. If it's a docker sdk `cartesi-machine` the path will be considered a path inside the docker sdk.

### DIFF
--- a/.changeset/salty-hoops-move.md
+++ b/.changeset/salty-hoops-move.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+fix ram_image default value
--- a/apps/cli/src/config.ts
+++ b/apps/cli/src/config.ts
@@ -73,7 +73,6 @@ export class InvalidStringArrayError extends Error {
  */
 const DEFAULT_FORMAT = "ext2";
 const DEFAULT_RAM = "128Mi";
-const DEFAULT_RAM_IMAGE = "/usr/share/cartesi-machine/images/linux.bin";
 export const DEFAULT_SDK_VERSION = "0.12.0-alpha.20";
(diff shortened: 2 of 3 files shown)


endersonmaia (approved):
> Note that if the user defines a ram_image in the cartesi.toml that path will be used. If it's a system cartesi-machine the path will be considered a host path. If it's a docker sdk cartesi-machine the path will be considered a path inside the docker sdk.

I think we could at least emit a warning if the ram_image is defined in the cartesi.toml and sdk's cartesi-machine is used.

The user can't inject a `linux.bin` inside the sdk easily.

I think the best approach would mount the ram_image into sdk container during build. This way, cartesi.toml ram_image path would always be the host's one.

I won't block, we can decide this later.

tuler:
> I think the best approach would mount the ram_image into sdk container during build. This way, cartesi.toml ram_image path would always be the host's one.

It can be complicated to mount, the docker desktop must have permission to mount the specific directory of the ram_image.
I think the use case for replacing the ram_image is an advanced one, and users who need that will likely have the cartesi-machine on the host, or have the experience to extend the sdk with a custom one.

PR #329, opened 2025-08-25
fix(cli): quote ENV vars when it contains spaces

endersonmaia:
This pull request addresses an issue with environment variable handling during Docker image builds in the CLI. The main improvement ensures that environment variable values containing spaces are properly quoted, preventing potential errors when passing them to cartesi-machine to boot.

**Environment variable handling improvements:**

* In `apps/cli/src/machine.ts`, updated the logic for constructing cartesi-machine `--env` arguments to detect environment variable values containing spaces and wrap those values in quotes. This ensures correct parsing by Docker when environment variables include spaces.

**Documentation:**

* Added a changeset file `.changeset/chilly-ducks-work.md` documenting this patch and describing the quoting of environment variables with spaces during build.

### DIFF
--- a/apps/cli/src/machine.ts
+++ b/apps/cli/src/machine.ts
@@ -45,7 +45,13 @@ export const bootMachine = (
 
     // list of environment variables of docker image
     const env = useDockerEnv ? (info?.env ?? []) : [];
-    const envs = env.map((variable) => `--env=${variable}`);
+    const envs = env.map((variable) => {
+        const eqIdx = variable.indexOf("=");
+        const key = variable.slice(0, eqIdx);
+        const value = variable.slice(eqIdx + 1);
+
+        return `--env=${key}="${value}"`;
+    });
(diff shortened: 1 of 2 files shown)


apps/cli/src/machine.ts
```diff
@@ -45,7 +45,13 @@ export const bootMachine = (
 
     // list of environment variables of docker image
     const env = useDockerEnv ? (info?.env ?? []) : [];
-    const envs = env.map((variable) => `--env=${variable}`);
+    const envs = env.map((variable) => {
+        const eqIdx = variable.indexOf("=");
```
tuler: I think this is a bit more elegant:

```typescript
    const envs = env.map((variable) => {
        const [key, value] = variable.split("=");
        return `--env=${key}="${value}"`;
    });
```

tuler (approved):
fixup

PR #331, opened 2025-09-06
feat(cli): add erc1155 to deposit commands

Nonnyjoe:
This PR adds support for single and batch ERC1155 deposits to applications via the CLI.

### DIFF
--- a/apps/cli/src/commands/deposit/erc1155.ts
+++ b/apps/cli/src/commands/deposit/erc1155.ts
@@ -1,0 +1,343 @@
+import { Command } from "@commander-js/extra-typings";
+import input from "@inquirer/input";
+import chalk from "chalk";
+import ora from "ora";
+import { type Address, getAddress, isAddress, isHex } from "viem";
+import { getProjectName } from "../../base.js";
+import {
+    erc1155BatchPortalAbi,
+    erc1155BatchPortalAddress,
+    erc1155SinglePortalAbi,
+    erc1155SinglePortalAddress,
+    testMultiTokenAbi,
+    testMultiTokenAddress,
+} from "../../contracts.js";
+import {
+    addressInput,
+    bigintInput,
+    getInputApplicationAddress,
+} from "../../prompts.js";
+import { connect } from "../../wallet.js";
+import type { DepositCommandOpts } from "../deposit.js";
+
+type ERC1155Token = {
+    address: Address;
+    name: string;
+};
+
+const parseToken = async (options: {
+    token?: string;
+}): Promise<ERC1155Token> => {
+    const address =
+        options.token && isAddress(options.token)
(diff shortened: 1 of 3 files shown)


apps/cli/src/commands/deposit/erc1155.ts
```diff
@@ -0,0 +1,339 @@
...
+            : await addressInput({
+                  message: "Token address",
+                  default: testMultiTokenAddress,
+              });
+
+    return { address, name: "TestMultiToken" };
+};
+
+export const createErc1155SingleCommand = () => {
+    return new Command<[], {}, DepositCommandOpts>("erc1155-single")
```
tuler: use `Record<string, never>` instead of `{}` to remove linter error

apps/cli/src/commands/deposit/erc1155.ts
```diff
@@ -0,0 +1,339 @@
...
+            const hash = await testClient.writeContract(request);
+            await testClient.waitForTransactionReceipt({ hash });
+            progress.succeed(
+                `Deposited ${amountLabel} to ${chalk.cyan(application)}`,
+            );
+        });
+};
+
+export const createErc1155BatchCommand = () => {
+    return new Command<[], {}, DepositCommandOpts>("erc1155-batch")
```
tuler: use Record<string, never> instead of {} to remove linter error

tuler:
Please squash both commits, and change the commit message to follow the repo other commits standard

tuler:
please change commit message to `"feat(cli): add erc1155 to deposit commands"`

apps/cli/src/prompts.ts
```diff
@@ -69,6 +69,22 @@ export const bigintInput = async (
     return parseUnits(value, config.decimals);
 };
 
+/**
+ * Prompt for a string value.
+ * @param config inquirer config
+ * @returns string
+ */
+export type StringPromptConfig = InputConfig & { default?: string };
```
tuler: I don't see this function being used. Can you remove it?

apps/cli/src/prompts.ts
```diff
@@ -69,6 +69,21 @@ export const bigintInput = async (
     return parseUnits(value, config.decimals);
 };
 
+/**
+ * Prompt for a string value.
+ * @param config inquirer config
+ * @returns string
+ */
+export const stringInput = async (
```
tuler: I don't see why we need a `stringInput` auxiliary function, as there is no special treatment being made here, just a plain call to `input`.
Can you remove the function and use `input` directly?

Nonnyjoe: Sure, I'll take out the stringInput function completely

tuler (approved):

PR #338, opened 2025-09-17
Bump dependencies

tuler:
Update all dependencies from workspace and CLI


endersonmaia:
Shouldn't the CI jog to run apps/cli tests be triggered?

If so, pleas adjust the ci so it runs the tests.

tuler:
> Shouldn't the CI jog to run apps/cli tests be triggered?

The `cli` workflow should run.

tuler:
> If so, pleas adjust the ci so it runs the tests.

Did a small change to `.github/workflows/cli.yaml`, and the workflow was now triggered.
I don't know though if the trigger was the change of the `.github/workflows/cli.yaml` file itself, which is included in the `paths` of the CI, or if the CI definitely needs quotes in paths, which is the change I made.

endersonmaia (approved):

PR #343, opened 2025-09-19
fix(cli): Add extra args to mksquashfs to deterministic files

lynoferraz:


endersonmaia (changes requested):
Add a changeset file and change the commit message to identify the cli module, likle `fix(cli): ...`.

tuler:
It's curious that `-all-time` is not enough and `-root-time` and -mkfs-time` are necessary.
From [documentation](https://manpages.debian.org/testing/squashfs-tools/mksquashfs.1.en.html):

```
[-all-time TIME](https://manpages.debian.org/testing/squashfs-tools/mksquashfs.1.en.html#all)
set all file timestamps to TIME. TIME can be an unsigned 32-bit int indicating seconds since the epoch (1970-01-01) or a string value which is passed to the "date" command to parse. Any string value which the date command recognises can be used such as "now", "last week", or "Wed Feb 15 21:02:39 GMT 2023".

[-root-time TIME](https://manpages.debian.org/testing/squashfs-tools/mksquashfs.1.en.html#root)
set root directory time to TIME. TIME can be an unsigned 32-bit int indicating seconds since the epoch (1970-01-01) or a string value which is passed to the "date" command to parse. Any string value which the date command recognises can be used such as "now", "last week", or "Wed Feb 15 21:02:39 GMT 2023".

[-mkfs-time TIME](https://manpages.debian.org/testing/squashfs-tools/mksquashfs.1.en.html#mkfs)
set filesystem creation timestamp to TIME. TIME can be an unsigned 32-bit int indicating seconds since the epoch (1970-01-01) or a string value which is passed to the "date" command to parse. Any string value which the date command recognises can be used such as "now", "last week", or "Wed Feb 15 21:02:39 GMT 2023".
```

tuler:
See new options for version 4.7

https://github.com/plougher/squashfs-tools/blob/master/CHANGES

endersonmaia:
> See new options for version 4.7
> 
> https://github.com/plougher/squashfs-tools/blob/master/CHANGES

The version inside `cartesi/sdk` image is `4.6.1`.

Version on my host machine is `4.7.2`.

endersonmaia (approved):

endersonmaia:
We already have `squashfs-tools` on its latest version on the sdk after PR #344 , and this already implements the necessary changes to make sqfs drives deterministic.

We could start using the new `-repro` flag added in 4.7.x version, but it's not a requirement.

> 	3. New options for building reproducible filesystems (Mksquash/Sqfstar)
> 
> 		3.1 Low level timestamp setting options extended:
> 			-mkfs-time inode sets filesystem creation time to the
> 			 latest inode timestamp
> 			-inode-time inode sets all inode timestamps to the
> 			 latest inode timestamp
> 			-root-time inode sets the root directory timestamp to
> 			 the latest inode timestamp
> 
> 		3.2 New easier to remember shorthand options
> 			-repro builds a reproducible filesystem image, it is
> 			 shorthand for -mkfs-time inode.
> 			-repro-time <time> builds a reproducible image, it is
> 			 shorthand for specifying -mkfs-time <time>
> 			 and -inode-time <time>.

tuler:
> We could start using the new `-repro` flag added in 4.7.x version, but it's not a requirement.

According to documentation `-mkfs-time` is not the same as `-repro`.
`-repro` is shorthand for specifying `-mkfs-time` and `-inode-time`.
Why not use `-repro`? Is it not needed to achieve reproducibility?

endersonmaia:
> Why not use -repro? Is it not needed to achieve reproducibility?

From my tests, this PR already achieves reproducibility.

tuler (approved):

PR #375, opened 2025-12-15
typescript compose builder

endersonmaia:
This pull request refactors the Docker Compose configuration for the CLI app by removing all static YAML Compose files and replacing them with TypeScript modules that define service and proxy configurations programmatically. Additionally, it consolidates environment variable management into TypeScript, updates dependencies, and simplifies the build process.

The most important changes are:

**Compose Configuration Refactor:**

[...]

PR #383, opened 2026-01-06
Changes to migrate compose from YAML to TS

tuler:
This changes the PR #375 in a few ways.

- it keeps the type system created for the compose structure.
- it creates a function to merge compose structures following the rules at https://docs.docker.com/reference/compose-file/merge/ (this was mostly AI generated).
- it changes from a class based generator to a simple concatenation of compose structures using the function above

I think this feels cleaner.
I did a basic test with basic services and explorer service. (account abstraction infrastructure is currently broken).


endersonmaia (changes requested):
While testing, I can't CTRL + C when I pick `(l) View logs`.

But the YAML part works as expected.

Tests were removed, I think the `concat()` functions deserves testing.

tuler:
> While testing, I can't CTRL + C when I pick `(l) View logs`.

Looks unrelated. Happens only on this branch?

endersonmaia:
> > While testing, I can't CTRL + C when I pick `(l) View logs`.
> 
> Looks unrelated. Happens only on this branch?

Didn't happen at `feature/ts-compose-builder` branch.

tuler:
> Didn't happen at `feature/ts-compose-builder` branch.

So we need to investigate this deeper

endersonmaia (approved):
Won't block for lack of unit tests for now.

PR #385, opened 2026-01-07
replace giget

endersonmaia:


apps/cli/src/template.ts
```diff
@@ -1,23 +1,68 @@
...
+    // Download the tarball
+    const response = await fetch(tarballUrl);
+    if (!response.ok) {
+        throw new Error(`Failed to download template: ${response.statusText}`);
+    }
+
+    // Stream download → gunzip → extract with filtering
+    // GitHub tarball structure: application-templates-{branch}/{template}/...
+    // We need to extract only files within the template subdirectory
+    const extractStream = unpackTar(out, {
```
tuler, line 32: I'd separate this in a function that unpacks a tar with a subdirectory option. The code below looks very specific to the template use case.

endersonmaia: but that's whatś  `unpackTar()` does, `out` is the dirertory where this will be unpacked

since we're inside a template.ts module, I didn't bother being specific here

tuler: I'm trying to simplify this code, but GitHub creates a directory inside the tar that is not exactly the branch name.

branch `prerelease/sdk-12` turns into directory `application-templates-prerelease-sdk-12` (`/` replaced by `-`). This is bad.

tuler (approved):

PR #386, opened 2026-01-08
Migration to bun

tuler:
This migrates from `pnpm` to `bun` as the package manager, and from `node` to `bun` as runtime platform.


tuler:
As far as I understand the release workflow is broken, because `changeset publish` is not compatible with `bun` yet, it only works with `npm` and `pnpm`.

We also need to understand how we are going to publish in case we opt for binaries produced by `bun build --compile`, as it's implemented right now. One idea is to use `brew` on Mac, and `apt-get` on Linux, but this still needs to be implemented.

endersonmaia:
> We also need to understand how we are going to publish in case we opt for binaries produced by `bun build --compile`, as it's implemented right now. One idea is to use `brew` on Mac, and `apt-get` on Linux, but this still needs to be implemented.

I think we could release the binaries via GitHub Releases at first, and think about brew and apt-get later.

tuler:
> I think we could release the binaries via GitHub Releases at first, and think about brew and apt-get later.

I changed the build process to do both for now.

1) produce `dist/index.js` as a minified bundle targeting nodejs. This should work as it is now, packaged and published to npmjs and executed with nodejs.
2) produce binaries at `bin/cartesi-${os}-${arch}`

I did not change the release workflow yet to save the binaries in the release. We need to see how to do that.

PR #387, opened 2026-01-08
Release with bun

tuler:
This simplifies a bit the release workflow and add the CLI binaries to the release


endersonmaia (approved):
I'm not able to build locally, let's mege ito to see if the CI runs the build jobs.

PR #401, opened 2026-01-13
Switching back chainId from 13370 to 31337 (SDK)

tuler:


endersonmaia (approved):
Couldn't test with CLI yet, but build works.

PR #402, opened 2026-01-13
Switching back chainId from 13370 to 31337 (CLI)

tuler:


endersonmaia (approved):
LGTM

Did some simple testes like `cartesi run, send, status, ....`

Maybe we could have proper unit/integration tests for this part of the code in the future.

PR #405, opened 2026-01-13
Use Bun.Archive API

endersonmaia:


tuler:
```typescript
const archiveFromFetch = new Bun.Archive(await response.blob());
```

Hmm, I think this is less efficient than the stream we are doing. The `await` will download the entire file to memory before handing over to Bun.Archive

Also, it does not seem to have an option to extract files stripping components (no map function on `ArchiveExtractOptions`)

endersonmaia:
Closing as the current Stream API is a better solution.

PR #410, opened 2026-01-14
chore(cli): bump minimum Docker and Docker Compose versions

endersonmaia:
`service.healthcheck.start_interval` is only available in Docker 25.0+ and
Compose 2.24.0+


tuler (approved):
deserves a changeset

PR #415, opened 2026-01-28
fix(cli): cpp template create

endersonmaia:
Creating a cpp template was fgailing beacuse it was conflicting with the cpp-low-level.

Made the match more restrict to the directory name.


tuler (commented):
Good catch, add changeset

tuler (approved):

PR #417, opened 2026-02-05
Support for ABI packed encoding at cartesi send

tuler:
This PR adds support for `cartesi send --encoding abi-packed`, to allow easy submission of inputs that use packed encoding. The following are valid commands:

```shell
cartesi-dev send --encoding abi-packed

cartesi-dev send --encoding abi-packed --abi-params uint8,uint64,uint16,uint64,int64,uint64,int64

cartesi-dev send --encoding abi-packed --abi-params uint8,uint64,uint16,uint64,int64,uint64,int64 3,1770308000,2,4412482,100000000000000000,4740165,3000000000000000
```

I also fixed the support for signed integers in the signature.


endersonmaia (approved):
LGTM

Exercising the command it seems to work as expected.

PR #418, opened 2026-02-10
Anvil polling interval

tuler:
The default polling interval of the viem connection to anvil is the default block time divided by 3, which is 12s / 3 = 4s, which is very long.
The default block time is already being set to 2s.
In a local test environment I think we can set the polling interval to a very low value, so the `send` operations are faster.
I'm setting here to 200ms, which makes the average time to send an input from more than 8s, to less than 2s.

### DIFF
--- a/.changeset/seven-hounds-teach.md
+++ b/.changeset/seven-hounds-teach.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+faster polling interval for cartesi send
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -34,10 +34,10 @@
         "ora": "^9.0.0",
         "p-retry": "^7.0.0",
         "progress-stream": "^2.0.0",
-        "semver": "^7.7.3",
+        "semver": "^7.7.4",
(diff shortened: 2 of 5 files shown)


brunomenezes (approved):
For DevNet, I think it's a no-brainer.  Would Anvil support websockets? I am not sure if that would be faster, though.

tuler:
> Would Anvil support websockets? I am not sure if that would be faster, though.

Anvil does support WebSocket, but I tried to use it right now and it didn't work.
It may be because we have a traefik proxy between the host and the docker anvil service, that may need special configuration for WebSocket.

I'll merge as is, and investigate that in the future.

tuler:
> It may be because we have a traefik proxy between the host and the docker anvil service, that may need special configuration for WebSocket.

According to traefik docs it should be automatic, so I don't think it's Traefik's fault.

> Traefik Proxy supports WebSocket (WS) and WebSocket Secure (WSS) connections out of the box. No special configuration is required beyond standard HTTP routing.

tuler:
Opened a separate draft for WebSocket connection
https://github.com/cartesi/cli/pull/419

PR #421, opened 2026-02-11
Add anvil fork mode to cartesi run

tuler:
## Summary

- Add `--fork` flag to `cartesi run` to fork from a live chain instead of using pre-baked devnet state
- Add `--fork-url <url>` option to specify a custom RPC endpoint (implies `--fork`, defaults to `https://ethereum.reth.rs/rpc`)
- Add `--fork-block-number <number>` option to pin the fork to a specific block

In fork mode, Anvil runs directly with `--fork-url` and `--chain-id 31337` instead of the `devnet` wrapper that loads `anvil_state.json`. All other services (node, explorer, bundler, paymaster) continue working unchanged since the chain ID remains 31337.

## Test plan

- [ ] `cartesi run --dry-run` — verify normal devnet mode is unchanged (no regression)
- [ ] `cartesi run --fork --dry-run` — verify anvil command uses fork with default URL
- [ ] `cartesi run --fork-url https://custom-rpc.example.com --dry-run` — verify fork with custom URL (implies `--fork`)
- [ ] `cartesi run --fork --fork-block-number 12345 --dry-run` — verify fork at specific block number
- [ ] `cartesi run --fork` — verify full environment starts and is healthy against a forked chain

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tuler:
Depending on the chosen fork-url the deployed contracts can have different addresses.
ps: in fact only 1 contract from Cartesi framework currently have this "problem": `DaveAppFactory`.

So the CLI needs to use the correct address during application deployment, and also report the correct address in the `address-book` command.

So it should know not only the addresses on local deployment, but also the addresses on all supported live networks, which can potentially be forked.

So I had to modify the devnet package to also package the deployments from all networks, which is maintained at cannon registry.

So I changed to devnet packaging to call `cannon inspect`, and export the addresses from all networks. And changed the artifact format to follow the format supported by the [wagmi-hardhat-plugin](https://github.com/sunodo/wagmi-plugin-hardhat-deploy), because then it would be much easier to use.

When merged the devnet package will need to be released first, then SDK, then CLI.

brunomenezes:
> ...
> So I changed to devnet packaging to call `cannon inspect`, and export the addresses from all networks. And changed the artifact format to follow the format supported by the [wagmi-hardhat-plugin](https://github.com/sunodo/wagmi-plugin-hardhat-deploy), because then it would be much easier to use.
> 
> When merged the devnet package will need to be released first, then SDK, then CLI.

So, are these devnet changes in some other PR? Also, I tried building the CLI from this branch, but it fails on the devnet package. Building it in isolation, here are the logs I received. Is there a specific version that is expected? Currently, my system is using 1.5.0-stable.

```txt
@cartesi/devnet:build:
@cartesi/devnet:build: IPFS Availability Score(# of nodes):  Run IPFS Locally to get this score
@cartesi/devnet:build:
✖ Failed to parse anvil version. Is anvil installed?
@cartesi/devnet:build: ✔ Download dependencies
@cartesi/devnet:build: ✔ Anvil started (PID: 98596)
@cartesi/devnet:build: ✔ Contracts deployed
@cartesi/devnet:build: ✔ Local deployments written to deployments/anvil.json
@cartesi/devnet:build: ✔ Export cannon packages
@cartesi/devnet:build: ✔ Anvil stopped -> anvil_state.json
@cartesi/devnet:build: error: script "build:dump" exited with code 1
@cartesi/devnet:build: ERROR: "build:dump" exited with 1.
@cartesi/devnet:build: error: script "build" exited with code 1
@cartesi/devnet:build: ERROR: command finished with error: command (/Users/--x--/cartesi/public/cli/packages/devnet) /opt/homebrew/bin/bun run build exited (1)
@cartesi/devnet#build: command (/Users/--x--/cartesi/public/cli/packages/devnet) /opt/homebrew/bin/bun run build exited (1)

 Tasks:    0 successful, 1 total                     └───────────────────┘
Cached:    0 cached, 1 total
  Time:    7.866s
Failed:    @cartesi/devnet#build
```

brunomenezes (approved):
I made the dry-runs suggested and also create/build/run an dapp. No hiccups.

tuler:
I think it's safer to remove the default fork url and require the user to provide one.

1. Removed `DEFAULT_FORK_URL` constant (line 31)
2. Removed the `--fork` option — the standalone flag that enabled fork mode with a default URL
3. Updated `configureFork` — removed the fork parameter; fork mode is now triggered only when `--fork-url` is provided (no default fallback)

Users who want to fork must now explicitly pass `--fork-url <url>`.

brunomenezes:
> I think it's safer to remove the default fork url and require the user to provide one.
> 
> 1. Removed `DEFAULT_FORK_URL` constant (line 31)
> 2. Removed the `--fork` option — the standalone flag that enabled fork mode with a default URL
> 3. Updated `configureFork` — removed the fork parameter; fork mode is now triggered only when `--fork-url` is provided (no default fallback)
> 
> Users who want to fork must now explicitly pass `--fork-url <url>`.

What was the motivation was because of rate limiting, because it goes straight to Ethereum-mainnet or a combination of both? Just picking your brain here. nonetheless I think that is better as a more deliberate decision.

apps/cli/src/base.ts
```diff
@@ -67,13 +67,22 @@ export type AddressBook = Record<string, Address>;
...
     projectName?: string;
 }): Promise<AddressBook> => {
+    const forkChainId = (await getForkChainId(options)) ?? 31337;
     const applicationAddress = await getApplicationAddress(options);
 
+    const chainDaveAppFactoryAddress =
+        daveAppFactoryAddress[
+            forkChainId as keyof typeof daveAppFactoryAddress
+        ];
+    if (!chainDaveAppFactoryAddress) {
```
brunomenezes, line 80: @tuler, should you do this kind of validation also on `compose/anvil.ts` and `compose/node.ts`? As this getAddressBook would only run when a *address-book* `command` is issued to the CLI. 

The validation is about whether the chain ID of the fork is supported or not, right?

tuler:
> What was the motivation was because of rate limiting

It was not working properly, the node was being disconnected often and dying.

apps/cli/src/base.ts
tuler: Good catch, added check as well to `node.ts`

tuler:
Improved behavior of address-book.
Now it prints a different set of contracts weather it’s a running fork or a running devnet.

Test tokens are only present at devnet.
ERC-4377 are currently only present at fork (not devnet).

endersonmaia:
When using an invalid URL it fails, I suggest catching the error and printing a nice message

```
cartesi run --dry-run --fork-url=https://url.com
WARNING: default block is set to 'latest', production configuration will likely use 'finalized'
i2 [HttpRequestError]: HTTP request failed.

URL: https://url.com/
Request body: {"method":"eth_blockNumber"}

Details: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
Version: viem@2.46.1
    at Object.request (file:///home/endersonmaia/Work/cartesi/cli/apps/cli/dist/index.js:288:28678)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    ... 3 lines matching cause stack trace ...
    at async X (file:///home/endersonmaia/Work/cartesi/cli/apps/cli/dist/index.js:288:3946) {
  details: `Unexpected token '<', "<!DOCTYPE "... is not valid JSON`,
  docsPath: undefined,
  metaMessages: [
    'URL: https://url.com/',
    'Request body: {"method":"eth_blockNumber"}'
  ],
  shortMessage: 'HTTP request failed.',
  version: '2.46.1',
  body: { method: 'eth_blockNumber', params: undefined },
  headers: undefined,
  status: undefined,
  url: 'https://url.com/',
  [cause]: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
      at JSON.parse (<anonymous>)
      at Object.request (file:///home/endersonmaia/Work/cartesi/cli/apps/cli/dist/index.js:288:28441)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async T (file:///home/endersonmaia/Work/cartesi/cli/apps/cli/dist/index.js:291:17428)
      at async request (file:///home/endersonmaia/Work/cartesi/cli/apps/cli/dist/index.js:291:17476)
      at async M3.delay.count (file:///home/endersonmaia/Work/cartesi/cli/apps/cli/dist/index.js:288:25384)
      at async X (file:///home/endersonmaia/Work/cartesi/cli/apps/cli/dist/index.js:288:3946)
}
```

endersonmaia:
I'm not sure how careful we should be with this kind or URL being accepted via CLI argument, since they may probably have some secret/token init that should be carefully handled. If we do not provide getting this from ENV we should at least show a disclaimer about rotating/revoking the token after using it.

tuler:
> I'm not sure how careful we should be with this kind or URL being accepted via CLI argument, since they may probably have some secret/token init that should be carefully handled. If we do not provide getting this from ENV we should at least show a disclaimer about rotating/revoking the token after using it.

The CLI is for dev environment, to run on the developer machine. I don't see a big risk here.

brunomenezes (approved):

PR #422, opened 2026-02-12
devnet live networks artifacts

tuler:
This PR adds to the devnet npm package information about the deployment of contracts to live networks (mainnets e testnets). This information is being fetched from the cannon registry.

The format of the artifacts also changed. Now each network has its own file under `deployments`. Each file contains all contracts, its ABIs and address. This format was inherited from `hardhat-deploy` and is ideal for offchain usage. As the format is the same as the `hardhat-deploy` one we can also use the [wagmi-plugin-hardhat-deploy](https://github.com/sunodo/wagmi-plugin-hardhat-deploy).

ps: also added here a CLAUDE.md to help with agentic coding.

### DIFF
--- a/packages/devnet/build.ts
+++ b/packages/devnet/build.ts
@@ -1,13 +1,35 @@
-import { semver } from "bun";
+import { $, semver } from "bun";
 import { existsSync, readdirSync } from "fs-extra";
 import { Listr, type ListrTask } from "listr2";
+import * as path from "node:path";
+import {
+    arbitrum,
+    arbitrumSepolia,
+    base,
+    baseSepolia,
+    mainnet,
+    optimism,
+    optimismSepolia,
+    sepolia,
+} from "viem/chains";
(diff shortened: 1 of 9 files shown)


packages/devnet/build.ts
```diff
@@ -1,13 +1,35 @@
...
+    optimism,
+    optimismSepolia,
+    sepolia,
+} from "viem/chains";
 import * as anvil from "./anvil";
 import { downloadAndExtract } from "./download";
-import path = require("node:path");
 
 const ANVIL_VERSION = "1.4.3";
 const ROLLUPS_VERSION = "2.1.1";
```
endersonmaia, line 19: Any reason why not using 2.2.0 already ?

tuler: Because the rollups-node is not using 2.2.0 yet.

endersonmaia (approved):

PR #423, opened 2026-02-12
correct eth_isready

tuler:
This PR bumps the devnet package, which has a different artifact structure. Some unused scripts were also deleted.
The `eth_isready` was fixed to optionally use a block number parameter that make it query the block number to signal that the service is ready. This will be useful to use a forked anvil.


packages/sdk/eth_isready
```diff
@@ -1,4 +1,13 @@
 #!/usr/bin/env sh
```
endersonmaia, line 1:
```suggestion
#!/usr/bin/env sh
set -eu
```

packages/sdk/devnet
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
```
endersonmaia:
```suggestion
#!/usr/bin/env sh
```

keep consistency between scripts

endersonmaia (approved):
Made little non blocking comments, but we should wait devnet release and CI get green.

PR #426, opened 2026-02-20
feat(cli): changes to run to work better with detached stdin (no shell)

tuler:
This makes `cartesi run` to work better with detached stdin, by not running the shell at the end.
You can test this by running `cartesi run < /dev/null`.
The goal is to make `run` work well when it's launched from another spawn.
The application deployment had to be moved from the shell loop to before the shell is called, so it happens with this new mode.

### DIFF
--- a/apps/cli/src/exec/rollups.ts
+++ b/apps/cli/src/exec/rollups.ts
@@ -135,7 +135,7 @@ type Service = {
     errorTitle?: string; // title of the service when it is not healthy
 };
 
-const host = "http://127.0.0.1";
+export const host = "http://127.0.0.1";
 
 // services configuration
 const baseServices: Service[] = [
@@ -242,6 +242,7 @@ export const startEnvironment = async (options: {
     blockTime: number;
     cpus?: number;
     defaultBlock: "latest" | "safe" | "pending" | "finalized";
+    detach: boolean;
     dryRun: boolean;
     forkConfig?: ForkConfig;
     memory?: number;
@@ -255,6 +256,7 @@ export const startEnvironment = async (options: {
         blockTime,
         cpus,
         defaultBlock,
+        detach,
         dryRun,
         forkConfig,
         memory,
@@ -265,8 +267,6 @@ export const startEnvironment = async (options: {
         verbose,
     } = options;
 
-    const address = `${host}:${port}`;
-
     // setup the environment variable used in docker compose
     const env: NodeJS.ProcessEnv = {
         CARTESI_BLOCKCHAIN_DEFAULT_BLOCK: defaultBlock,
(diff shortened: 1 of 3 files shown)


endersonmaia (changes requested):
for me Anvil never gets healthy 

```
docker inspect --format "{{json .State.Health.Log }}" my-go-dapp-anvil-1 | jq .
...
    "ExitCode": 2,
    "Output": "/usr/local/bin/eth_isready: 7: 1: parameter not set\n"
...
```

It was introduced when defining `set -u` in the `eth_isready` in cartesi/sdk.

An easy fix should be something like 

```shell
BLOCK_NUMBER=${1:-}

if [ -n "$BLOCK_NUMBER" ]; then
...

**Comment by tuler** [comment]
> for me Anvil never gets healthy
> 
> ```
> docker inspect --format "{{json .State.Health.Log }}" my-go-dapp-anvil-1 | jq .
> ...
>     "ExitCode": 2,
>     "Output": "/usr/local/bin/eth_isready: 7: 1: parameter not set\n"
> ...
> ```
> 
> It was introduced when defining `set -u` in the `eth_isready` in cartesi/sdk.
> 
> An easy fix should be something like
> 
> ```shell
> BLOCK_NUMBER=${1:-}
> 
> if [ -n "$BLOCK_NUMBER" ]; then
> ...
> ```

Hmm, that's for another PR right? How did it work before?

**Comment by tuler** [comment]
Maybe we should rethink how to do this better, because today docker compose is always executed in the background with `--detach`. But maybe in the scenario where TTY is not there, and shell is not there, it should run on foreground.

**Comment by endersonmaia** [comment]
> Hmm, that's for another PR right?

I'll send that PR. 

> How did it work before?

We're focusing on testing the `--fork-url`, and for that, it works since we have a block number.

It fails without `--fork-url`.

**Comment by tuler** [comment]
Changed the code to not run compose detached if TTY is not attached

File: apps/cli/src/exec/rollups.ts [hunk 1]
```diff
@@ -330,25 +333,21 @@ export const startEnvironment = async (options: {
...
 
-    // pull images first
-    // const pullArgs = ["--policy", "missing"];
-    // await execa("docker", [...composeArgs, "pull", ...pullArgs], {
-    //     env,
-    ////FIXME: stdio and input won't work together
-    //     stdio: "inherit",
-    //     input: composeFile.build()
-    // });
-
```
endersonmaia, line 344: This was commented when migrated from YAML to TS, and compose file being passed via `stdin`.

When executed for the first time, it downloads on the backgound but it doesn' t gives any feedback to the user, looks like it's stuck.

We need a solution for this, but won't blok this PR for that.

endersonmaia (approved):

PR #448, opened 2026-03-11
feat(sdk): add telegraf to rollups-runtime

endersonmaia:
When deploying a rollups-node using the cartesi/rollups-runtime container image, it's usefull to have some way to collect metrics about the environment.

I propose we add telegraf as an option inside the image, just like we have other auxiliary tools to supervise the services like `nitro` and `tini`.

Those who will use telegraf, should be responsible to add the necessary configuration.

From the Cartesi Infra side, we're already using and having to add telegraf everywhere, so having this as part of the runtime image would save some time.

### DIFF
--- a/packages/sdk/Dockerfile
+++ b/packages/sdk/Dockerfile
@@ -118,6 +118,25 @@ tini --version
 EOF
 
 ################################################################################
+# install telegraf
+FROM base AS telegraf
+ARG TELEGRAF_VERSION
+ARG TARGETARCH
+WORKDIR /tmp
+RUN <<EOF
+curl -fsSL https://dl.influxdata.com/telegraf/releases/telegraf-${TELEGRAF_VERSION}_linux_${TARGETARCH}.tar.gz \
+    -o ./telegraf.tar.gz;
+case "${TARGETARCH}" in
+    amd64) echo "57e6d733de44335127c06d9d9099c9dee7307beff992783438ce38b9ba1b8e5e /tmp/telegraf.tar.gz" | sha256sum --check ;;
+    arm64) echo "d42aa06b113f15fdc14775a2d1971e31a0adb82a6e0da28e40a17e819ddaeb06 /tmp/telegraf.tar.gz" | sha256sum --check ;;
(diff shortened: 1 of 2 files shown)


packages/sdk/Dockerfile
```diff
@@ -117,6 +117,25 @@ chmod +x /usr/local/bin/tini
...
+RUN <<EOF
+curl -fsSL https://dl.influxdata.com/telegraf/releases/telegraf-${TELEGRAF_VERSION}_linux_${TARGETARCH}.tar.gz \
+    -o ./telegraf.tar.gz;
+case "${TARGETARCH}" in
+    amd64) echo "57e6d733de44335127c06d9d9099c9dee7307beff992783438ce38b9ba1b8e5e /tmp/telegraf.tar.gz" | sha256sum --check ;;
+    arm64) echo "d42aa06b113f15fdc14775a2d1971e31a0adb82a6e0da28e40a17e819ddaeb06 /tmp/telegraf.tar.gz" | sha256sum --check ;;
+    *) echo "unsupported architecture: ${TARGETARCH}"; exit 1 ;;
+esac
+tar -xvzf /tmp/telegraf.tar.gz -C / --strip-components=2 ./telegraf-${TELEGRAF_VERSION}/usr/bin/telegraf
+cp /usr/bin/telegraf /usr/local/bin/telegraf
```
mpolitzer, line 135: Why move to `/usr/local` and copy back to the original place later?

mpolitzer (commented):
LGTM

What is the idea for telegraf configs, are they coming from somewhere else, not needed?

packages/sdk/Dockerfile
endersonmaia: OCD, in the final image I want a `COPY ... /usr/local/bin/ ...` for every `COPY` line.

endersonmaia:
> What is the idea for telegraf configs, are they coming from somewhere else, not needed?

It's up to the user/operator of the node to inject the necessary config. I don't want to suggest anything.

mpolitzer (approved):

lynoferraz (approved):

PR #449, opened 2026-03-12
chore(sdk): Bump devnet and rollups-node version.

brunomenezes:
# Summary 
This change is upgrading the devnet and rollups-node versions in the sdk package.

## Dependencies

- [x] Rollups node alpha.10 release.

Once the dependency is resolved, I will amend the initial commit to include the full set of changes and update the PR to ready-to-review.

PR #450, opened 2026-03-13
feat: upgrade sdk, devnet and rollups-node npm package.

brunomenezes:
# Summary
code change to support the upcoming node `alpha.10`.  The changes are related to bumping devnet to `v2.0.0-alpha.11`, @rollups/contracts `v2.2.0` and  the SDK package.

The artefact used for this development: build artefact can be found [here](https://github.com/cartesi/rollups-node/actions/runs/22962919996) from branch `next/2.0`. 

## Dependencies

- [x] Merge of PR #449 

The dependency waits on the release of the node `alpha.10`. Once it is done, I can update the `sdk` and release it, then update this PR accordingly, and once it is merged, a new CLI release can be done.

### DIFF
--- a/.changeset/every-spies-exist.md
+++ b/.changeset/every-spies-exist.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+Replace --force flag to the new --yes to skip confirmation prompts on apps remove action.
--- a/.changeset/smooth-animals-write.md
+++ b/.changeset/smooth-animals-write.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+Bump devnet@2.0.0-alpha.11 and cartesi/rollup@2.2.0 npm packages.
--- a/.changeset/smooth-years-jam.md
+++ b/.changeset/smooth-years-jam.md
@@ -1,0 +1,5 @@
+---
+"@cartesi/cli": patch
+---
+
+Bump CLI default SDK version from alpha.34 to alpha.37.
--- a/apps/cli/package.json
+++ b/apps/cli/package.json
@@ -41,8 +41,8 @@
         "yaml": "^2.8.2"
     },
     "devDependencies": {
-        "@cartesi/devnet": "2.0.0-alpha.10",
-        "@cartesi/rollups": "2.1.1",
+        "@cartesi/devnet": "2.0.0-alpha.11",
+        "@cartesi/rollups": "2.2.0",
         "@sunodo/wagmi-plugin-hardhat-deploy": "^0.4.0",
         "@types/bun": "^1.3.6",
         "@types/bytes": "^3.1.5",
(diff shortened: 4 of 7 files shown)


brunomenezes:
This commit here 827b23e1cb1ab1bb6007c132c40cbbf7d8db90a4 will be removed once the PR #449 is merged, as I pushed that change there too. So, if you use `cartesi run`, the change avoids issues.

brunomenezes:
I recorded a little demo. (UPDATE: Ignore as the `apps` command was removed) 

https://github.com/user-attachments/assets/765b19e1-ace5-4e56-bf44-38b16ff9a2c2

tuler:
Why do we need this new command?

brunomenezes:
> Why do we need this new command?

As I understand it, a user can, at ease, re-enable an application that went into a `failed` state. Let me know if you don't think the change for the upcoming rollups-node does not worth the maintenance, and I will undo it.

tuler:
> As I understand it, a user can, at ease, re-enable an application that went into a `failed` state. Let me know if you don't think the change for the upcoming rollups-node does not worth the maintenance, and I will undo it.

From an node operator perspective it makes sense, but from an application developer perspective, why does he needs that command?

brunomenezes:
> > As I understand it, a user can, at ease, re-enable an application that went into a `failed` state. Let me know if you don't think the change for the upcoming rollups-node does not worth the maintenance, and I will undo it.
> 
> From an node operator perspective it makes sense, but from an application developer perspective, why does he needs that command?

What are the profile or profiles the cartesi-cli  was meant to support? Is the initial assumption that it is only for application developers? Also, who develop the application never operates it?

tuler:
> What are the profile or profiles the cartesi-cli was meant to support?

Application developers developing locally.
For application operation on testnet or mainnet the rollups-node is used directly.

brunomenezes:
So, by what @tuler is saying, I figure we don't need that in the cartesi-cli for the users we discussed earlier @vfusco , do you agree with that? They would do it directly using the rollups-node. Also, it is my understanding that the release is just to package the next node, since the current set of features is perfectly fine without any extras, right @tuler? Just making sure my understanding of this thread is sound.

packages/sdk/eth_isready
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
```
tuler: this will only go into a new SDK release if there is a changeset for it

brunomenezes: I mentioned [here](https://github.com/cartesi/cli/pull/450#issuecomment-4053364645) that the commit will be removed as I already pushed one in #449 , here is just in case one of you wanted to have a run and test locally to avoid figuring out why rollups-node was not starting.

vfusco (approved):

endersonmaia (approved):

PR #464, opened 2026-04-21
ci: bump docker actions

brunomenezes:
# Summary
Group change targeting GH actions.

After merging this PR will be able to close the following, which is targeting `main` rather than `prerelease/v2-alpha`

* #440 
* #441 
* #442 
* #443


endersonmaia (approved):
Changes seems ok, but it's curious why the `sdk.yaml` workflow didn't run.

PR #485, opened 2026-06-03
feat(cli): add --cache-from and --cache-to options to cartesi build

### LINKED ISSUE #483: Docker cache support
## 📚 Context

`cartesi build` uses `docker buildx build` under the hood when building a `docker` drive.
Buildx support a powerful configurable cache system, including cache support when [running in GitHub Actions](https://docs.docker.com/build/cache/backends/gha/).

`cartesi build` does not expose options to configure caching.
Caching would greatly improve build times of `cartesi build` in CI.

## ✔️ Solution

Allow a configuration at `cartesi.toml` docker drive to set `--cache-to` and `--cache-from`.

The current possible workaround is to build a docker image without using `cartesi build`, leveraging cache, and then using `cartesi build` with a pre-existing docker image with the `image` toml option.

endersonmaia:
Expose Docker Buildx cache backend specs via two new repeatable CLI flags:

  cartesi build --cache-from type=local,src=/tmp/cache \
                --cache-to   type=local,dest=/tmp/cache

The same options are also available per-drive in cartesi.toml:

  [drives.root]
  cache_from = ["type=gha"]
  cache_to   = ["type=gha,mode=max"]

CLI flags override the TOML values when non-empty.

Changes:
- config.ts: add cacheFrom/cacheTo fields to DockerDriveConfig and
  defaultRootDriveConfig; parse cache_from/cache_to from TOML
- builder/docker.ts: thread cacheFrom/cacheTo through ImageBuildOptions
  and append --cache-from / --cache-to args to the buildx invocation
- commands/build.ts: add --cache-from and --cache-to options (accumulator
  style); merge into BuildContext and apply over per-drive config
- tests/unit/config.test.ts: add assertion for cache_from/cache_to parsing
- tests/integration/builder/docker.test.ts: add two integration tests —
  one verifying the local cache directory is written, one verifying that
  layers are recovered from a local cache after the daemon build cache is
  pruned (attested by comparing root.ext2 SHA-256 between the two builds)
- tests/integration/builder/fixtures/Dockerfile.cache: new fixture using
  a non-deterministic RUN date layer to make cache-hit vs miss detectable

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

/close #483

### DIFF
--- a/apps/cli/src/commands/build.ts
+++ b/apps/cli/src/commands/build.ts
@@ -17,6 +17,8 @@ import { bootMachine } from "../machine.js";
 
 // context for Listr build tasks
 interface BuildContext {
+    cacheFrom: string[];
+    cacheTo: string[];
     config: Config;
     debug: boolean;
     destination: string;
@@ -37,6 +39,8 @@ const buildDriveTask = (
                 break;
             }
             case "docker": {
+                if (ctx.cacheFrom.length > 0) drive.cacheFrom = ctx.cacheFrom;
+                if (ctx.cacheTo.length > 0) drive.cacheTo = ctx.cacheTo;
                 const imageInfo = await buildDocker(
                     name,
                     drive,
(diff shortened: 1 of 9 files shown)


apps/cli/src/commands/build.ts
```diff
@@ -37,9 +39,16 @@ const buildDriveTask = (
                 break;
             }
             case "docker": {
+                const effectiveDrive = {
```
tuler: this code block is a little ugly

endersonmaia: Take a look at 1790c3a and see if it's better now.

tuler (approved):
Looks good. Fixup

PR #486, opened 2026-06-10
feat(devnet): Deploy withdrawal output builder

brunomenezes:
# Summary
This PR introduces a dedicated, deterministic deployment script for `UsdWithdrawalOutputBuilder`, making it available in the deployment and final anvil state dump (`anvil_state.json`) deliverables (devnet only, 31337). It will support the usage and testing of emergency withdrawal. Also, to avoid our own inline implementation of its `UsdWithdrawalOutputBuilderFactory` interface and avoid misalignments with subsequent changes in the rollups-contracts code, the `cartesi-rollups-contracts` was added as a soldeer dependency so the Factory definition can be used in the script.

PS: In future rollups-contracts releases, this contract will be deployed and available. Therefore, the script will be removed, and only the content from the artifacts will be necessary. Nonetheless, it will be transparent for clients using this package. 

### What Changed:
* Add forge script in the build pipeline to deploy `UsdWithdrawalOutputBuilder` to be included in the deployments and anvil state dump. 
* Removed its own implementation of test tokens. Therefore, only those from `rollups-contracts` will be part of the distribution. **Clients will need to update accordingly  (e.g. cartesi-cli address-book)**


packages/devnet/script/DeployUsdWithdrawalOutputBuilder.s.sol
```diff
@@ -0,0 +1,40 @@
...
+
+        // Keep this script idempotent when running against a state dump that
+        // may already contain this token+salt deployment.
+        if (builder.code.length == 0) {
+            address deployed = address(factory.newUsdWithdrawalOutputBuilder(token, SALT));
+            vm.assertEq(deployed, builder);
+            builder = deployed;
+        }
+
+        _storeDeployment("UsdWithdrawalOutputBuilder", builder);
```
guidanoli: I would clearly label the contract with a `Test` prefix, in order to avoid mistakes in production, where developers use this contract address when deploying applications to real networks. It's cheap insurance for a catastrophic event (users wouldn't be able to withdraw).

```suggestion
        _storeDeployment("TestUsdWithdrawalOutputBuilder", builder);
```

guidanoli (changes requested):

packages/devnet/script/DeployUsdWithdrawalOutputBuilder.s.sol
brunomenezes: @guidanoli, 
I think the ROI is really small here. Nonetheless, here is the change to support the name change [090e14c](https://github.com/cartesi/cli/pull/486/commits/090e14c78a3bf8589b41e41c567be5889cf4033f).

endersonmaia (approved):
LGTM

PR #489, opened 2026-06-15
feat: changes to support node alpha.12

### LINKED ISSUE #473: CLI: Bump SDK version v0.12.0-alpha.40 [next]
## 📚 Context
Update the CLI config to use the Cartesi/SDK image version `0.12.0-alpha.40`

## Dependencies

- [ ] Depends  on image publishing [cartesi/sdk dockerhub](https://hub.docker.com/r/cartesi/sdk/tags) 🟧

This image release should include the following changes: 
* #470 
* #471 
* #472 

## 📈 Subtasks

-   [ ] Update `DEFAULT_SDK_VERSION` property on `config.ts` file.
-   [ ] Verify whether everything is working as expected. i.e. build and use a local CLI, docker-desktop should download and use `cartesi/sdk:0.12.0-alpha.40` to build a simple cartesi application with success.

### LINKED ISSUE #474: CLI: Bump devnet version 2.0.0-alpha.12 [next]
## 📚 Context
Update the CLI dependency `@cartesi/devnet` to version `2.0.0-alpha.12`

## Dependencies

- [ ] Depends  on package publish to npm  [@cartesi/devnet](https://www.npmjs.com/package/@cartesi/devnet?activeTab=versions) 🟧

The release should include the following changes: 
* #465  
* #466 
* #467 

## 📈 Subtasks

-   [ ] Bump `@cartesi/devnet` dependency to version 2.0.0-alpha.12
-   [ ] Verify whether everything is working as expected.

### LINKED ISSUE #476: Fix: Non existing cartesi-machine option
## 🙂 Expected behavior

CLI should not add in the arguments options flags that is non-existing in the cartesi-machine.

## 🫠 Actual behavior

There is a `no-rollup` available to configure. It is passed down when running the `cartesi build` command.

## 🧪 Minimal test case

using the latest CLI (alpha.34) and having cartesi-machine installed or at least Docker available.

1. Create a simple Application `cartesi create --template=<template_name> <app_name>`
2. `cd <app_name>`
3.  create a `cartesi.toml` file 
4. add the following configuration 

```toml
[machine]
no_rollup=true
```
5. Run `cartesi build`

Code ref: https://github.com/cartesi/cli/blob/prerelease/v2-alpha/apps/cli/src/machine.ts#L111
`cartesi.toml` test ref: https://github.com/cartesi/cli/blob/prerelease/v2-alpha/apps/cli/tests/unit/config/fixtures/full.toml#L9

## ✔️ Possible solutions

1. If it is a mistake remove that option from both code and test references
2. if it is a typo due to a `--no-rollback` exist, replace accordingly.

### LINKED ISSUE #484: CLI: Update commands that user cartesi-rollups-cli
## 📚 Context

Upcoming rollups-node alpha.12 will have changes to some of the commands used by its `cartesi-rollups-cli`, such as `deploy` and `apps`, which are internally used by the `cartesi-cli`. So updates are needed to the input when calling these functions, and also for handling the returned data.  

rollups-node changes ref: https://github.com/cartesi/rollups-node/pull/779

## ✔️ Solution
Update the CLI to add support for the new available parameters to the aforementioned commands. 

## 📈 Subtasks

- [x] Update deploy parameters command.
- [x] Update deploy return information.
- [x] Update app [list, status, remove] commands.
- [x] Update app [list, status, remove] return.

### LINKED ISSUE #488: Forward host environment variables to the rollups node container
## 📚 Context

The rollups node supports fine-grained configuration through environment variables that the CLI does not expose as flags. For example, `CARTESI_LOG_LEVEL_JSONRPC_API` controls the log level of the JSON-RPC API subsystem independently of the global `CARTESI_LOG_LEVEL`.

Today there is no way to set such a variable through the CLI. Defining it in the shell when running `cartesi run` has no effect:

```shell
CARTESI_LOG_LEVEL_JSONRPC_API=debug cartesi run
# variable is not present in the rollups_node container
```

The reason is that the environment passed to `docker compose` (in `startEnvironment` at `apps/cli/src/exec/rollups.ts`) only reaches the compose **client process** — execa already merges `process.env` by default (`extendEnv: true`). Docker Compose does not forward its own process environment into containers: a container's environment comes solely from the `environment:` / `env_file:` sections of the compose file. The host-side environment is only used for `${VAR}` interpolation in the YAML and for valueless passthrough keys (e.g. `- FOO`). The compose services generated by the CLI (e.g. `apps/cli/src/compose/node.ts`) define fully literal `environment:` maps, with no interpolation and no passthrough keys — so nothing from the user's shell can reach the containers.

## ✔️ Solution

Possible approaches (not mutually exclusive):

1. **Prefix passthrough**: in `startEnvironment`, collect all `CARTESI_*` variables from `process.env` and inject them into the `rollups_node` service `environment:` section.
    - Benefit: zero ceremony, works for any current and future node variable.
    - Drawback: ambient shell variables silently affect the node; precedence must be defined (should a shell `CARTESI_LOG_LEVEL` override the value derived from `--verbose`?).
2. **Explicit flag**: add `--env KEY=VALUE` (repeatable) to `cartesi run`.
    - Benefit: explicit and discoverable, no ambient-environment surprises.
    - Drawback: more typing for something the user already expressed as an env var.
3. **Project config**: support an env section in `cartesi.toml` for persistent per-project node configuration.
    - Benefit: declarative and versionable.
    - Drawback: not suited for one-off overrides.

## 📈 Subtasks

- [x] Decide approach and precedence rules (CLI-flag-derived values vs user-provided values)
- [x] Implement forwarding in `startEnvironment` / compose generation
- [ ] Add tests covering the generated compose output (`--dry-run` config)
- [ ] Document the behavior

## 🎯 Definition of Done

- [ ] `CARTESI_LOG_LEVEL_JSONRPC_API=debug cartesi run` (or the equivalent flag/config) results in the variable being set inside the `rollups_node` container

brunomenezes:
# 📝 Summary
This PR improves the CLI run flow, node compose generation, and configuration parsing, alongside dependency upgrades, test suite expansions, and comprehensive validation for fork mode configurations.

## 🚀 Key Changes
* **CLI & run Command Enhancements**
    * Added a flag to list supported environment variables per service in JSON format.
    * Added support for a claim staging period in the run/deploy flow.
    * Added support for loading and passing withdrawal configurations via TOML files.
    * Enabled passing direct configuration file paths to the run command.
    * Added tty Docker option support in the fallback execution path (used by the shell command).
    * Fixed a bug by removing the `--no-rollup` flag passed to the `cartesi-machine` binary, ensuring compatibility with newer versions.
    * Integrated fork configuration assertions in `run.ts` before initiating a run session in fork mode.

* **Config System & Validation**
    * **Verification of Required Settings on Fork Mode:** Created a new validation module `validations.ts` with `assertForkConfig` to execute pre-flight assertions. This ensures required contracts are deployed on the target chain/block and validates that on-chain calls like `getInputBox` or `getDeploymentBlockNumber` succeed.
    * Replaced `getForkChainId` with a more complete `getForkConfig` function inside `rollups.ts` to query the current Anvil service details and return a complete `ForkConfig` (containing `chainId`, `url`, and `blockNumber`).
    * Integrated validations into `base.ts` to verify the deployment and contract state within `getAddressBook()`.
    * Introduced `withdrawalConfig` parsing and validation (including address and numeric validation).
    * Added `InvalidAddressValueError`, `ForkChainValidationError`, and `UnsupportedForkChainError` to improve error context.
    * Updated machine hash extraction to read from `hash_tree.sht` (root hash offset).

* **Rollups-Node Env Management & Proxy Updates**
    * Implemented a strict allow-list for CARTESI-prefixed environment variables. Host-provided values take precedence over defaults set in the code.
    * Removed `CARTESI_BLOCKCHAIN_WS_ENDPOINT` from default node compose environments.
    * Added CORS middleware (wildcard origins, methods, and headers) definition to the node proxy's inspect and RPC routers.

* **Dependencies & SDK**
    * Bumped default SDK to `0.12.0-alpha.41`.
    * Bumped the `rollups-explorer` image integration version from `2.0.0-alpha.2` to `2.0.0-alpha.3` in `rollups.ts`.
    * Updated devnet dependencies and removed unused rollups dependencies.
    * Updated address book entries for renamed token constants and integrated `TestUsdWithdrawalOutputBuilder`.

### ⚠️ Scope & Impact (Note for Reviewers)

Migration risk is low, but note the following behavioral changes:

1. **Environment Filtering:** Unsupported environment variables are now explicitly filtered out by the new allow-list.
2. **Machine Hash Source:** The machine hash source has shifted to `hash_tree.sht`. (Don't forget to execute `cartesi build` on existing dapps)
3. **Flag Removal:** The `--no-rollup` flag has been removed as it is not available as an option in `cartesi-machine` versions `0.20` and previous `0.19`. This ensures the CLI works correctly with newer CM versions.
4. **Fork Mode Safety:** Fork-mode runs will now fail early if pre-flight assertions on contract bytecode or vital on-chain calls (like `getInputBox`) fail.

### 🧪 Testing & Validation

* **Unit Testing:** Added comprehensive test cases for `validations.ts` that dynamically derive contract baseline call indices (avoiding fragile, hardcoded call-count assumptions) and explicitly assert contract checks.
* Expanded unit tests covering node compose allow-lists, default build behaviors, and withdrawal config parsing.


brunomenezes:
I am making that ready to review, but a release will also need to have an explorer ready 

The PR is here: https://github.com/cartesi/rollups-explorer/pull/464. 

Once we have a tag and release for the explorer, I will just update the image tag for the service.

apps/cli/src/commands/run.ts
```diff
@@ -321,12 +327,25 @@ export const createRunCommand = () => {
...
 
+            if (listSupportedVariables) {
+                const allowedVarsByService = {
+                    rollupsNode: nodeAllowedEnvironmentVariables,
+                };
+
+                // output the allowed environment variables by service in a JSON format and quit
+                process.stdout.write(
+                    JSON.stringify(allowedVarsByService, null, 2),
+                );
```
endersonmaia, line 341: Instead of printing the json, print only the variables, it's cleaner and conssistend with other outputs like `cartesi address-book`

```
cartesi run --list-supported-variables
CARTESI_AUTH_MNEMONIC
CARTESI_AUTH_MNEMONIC_ACCOUNT_INDEX
CARTESI_BLOCKCHAIN_DEFAULT_BLOCK
CARTESI_BLOCKCHAIN_HTTP_AUTHORIZATION
CARTESI_BLOCKCHAIN_HTTP_ENDPOINT
CARTESI_BLOCKCHAIN_ID
CARTESI_CONTRACTS_APPLICATION_FACTORY_ADDRESS
CARTESI_CONTRACTS_AUTHORITY_FACTORY_ADDRESS
CARTESI_CONTRACTS_DAVE_APP_FACTORY_ADDRESS
CARTESI_CONTRACTS_INPUT_BOX_ADDRESS
CARTESI_CONTRACTS_QUORUM_FACTORY_ADDRESS
CARTESI_CONTRACTS_SELF_HOSTED_APPLICATION_FACTORY_ADDRESS
CARTESI_DATABASE_CONNECTION
CARTESI_LOG_LEVEL
CARTESI_LOG_LEVEL_ADVANCER
CARTESI_LOG_LEVEL_CLAIMER
CARTESI_LOG_LEVEL_EVM_READER
CARTESI_LOG_LEVEL_JSONRPC_API
CARTESI_LOG_LEVEL_PRT
CARTESI_LOG_LEVEL_VALIDATOR
CARTESI_JSONRPC_MACHINE_LOG_LEVEL
CARTESI_SNAPSHOTS_DIR
```

endersonmaia (changes requested):
I see there's `--prt` that enables prt mode and default is Authority, what about Quorum, will it be possible to use Quorum in dev mode?

apps/cli/src/commands/run.ts
brunomenezes: The reasoning behind the choice is to highlight that this is a list of allowed variables for the rollups-node service, since the only part concerned with it is within `compose/node`. 

The `address-book` prints a table-formatted content. The `--dry-run` prints a `yaml formatted` configuration so not much of a consistent format, but I guess it has intention and makes sense how they produce the output.

brunomenezes:
> I see there's `--prt` that enables prt mode and default is Authority, what about Quorum, will it be possible to use Quorum in dev mode?

Not for this release.

brunomenezes:
I did one extra commit to include a fix to the following bug

- #476

apps/cli/src/compose/node.ts
```diff
@@ -19,6 +20,69 @@ type ServiceOptions = {
...
+    cartesiEnvironmentVariables?: CartesiEnvironmentVariables;
+};
+
+/**
+ * A list of allowed environment variables
+ * that can be passed to the rollups node service.
+ * A number of environment variables are rule out to avoid confusion e.g. CORS, FEATURE Enablement, etc.
+ */
+export const nodeAllowedEnvironmentVariables = [
+    "CARTESI_AUTH_MNEMONIC",
```
tuler, line 32: is it really worth the burden of maintaining this list over time?

apps/cli/src/base.ts
```diff
@@ -22,24 +23,33 @@ import {
...
+import { assertForkConfig } from "./validations.js";
 
 export const getContextPath = (...paths: string[]): string => {
     return path.join(".cartesi", ...paths);
 };
 
 export const getMachineHash = (): Hash | undefined => {
     // read hash of the cartesi machine snapshot, if one exists
-    const hashPath = getContextPath("image", "hash");
+    const hashPath = getContextPath("image", "hash_tree.sht");
```
tuler: I wouldn't read this file. Instead I'd rather call the `cartesi-machine-stored-hash` utility

brunomenezes: Push the change here: https://github.com/cartesi/cli/pull/489/changes/053f405ab2eebdd899de6d258a872397965553b3

The change in the YAML was small, just the QEMU + Docker Buildx setup. ~~But for some reason, Biome was formatting the whole document.~~

apps/cli/src/compose/node.ts
brunomenezes: I did curate that because there are a dozen more; I just did not add them all because, looking at the node source code, it looks like they set reasonable defaults. I am not sure about going free and just plugging any `CARTESI_*` in the compose environment section of the rollups-node service declaration.

tuler (approved):

apps/cli/src/compose/node.ts
tuler: I think only advanced users will set `CARTESI_` variables. And for those it's a minor issue if they break the behavior of the dev execution because they set some variable to some value that changes the behavior in an unpredictable way to the CLI.

IMO it's a big downside for maintenance to keep this list. The rollups node may change variables, as it happened several times in the past, and the CLI will always need to keep up.

I'd just iterate over all variables, filter all prefixed with `CARTESI_` and forward all. It's simple, it works, and low-maintenance.

Issue #477, opened 2026-04-30
New Options to cartesi.toml


lynoferraz:
While working on cartesapp framework I added some extra features that improved the experience and I think could easily be added in the cli too. All the features described here are controlled via the cartesi.toml file.

### `[machine]` block
- **workdir** = "path": Sets the workdir on the cartesi-machine cli wiht `--workdir=path`. Overrides docker `WORKDIR` command.
- **entrypoint** = "command": Sets the entrypoint on the cartesi-machine cli wiht `-- command`. Overrides docker `ENTRYPOINT` and `CMD` commands.
- **envs** = ["env1=val1",...]:  Additional envs to be set in the env
- **network** = "true": enable network to customize the rootfs in shell mode.
- **no_bootargs** = "true|false": enable option.
- **no_rollup** = "true|false": enable option.
- **init** = ["cmd1","cmd2",...]:  Additional `--append-init` options.
- **bootargs** = ["cmd1","cmd2",...]:  Additional `--append-bootargs` options.

### `[drive]` block

for `format = "ext2"`
- **size** = "size str": use `xgenext2fs` `--size-in-blocks=blocks` option (it should convert first)

for `builder = "docker"`.
- **format** = "none":  Allows an unchanged format. This allows to create files in a dockerfile step and used it unchanged as a flash drive. Useful for creating state files or drives directly in a docker step.
- **output_type** = "none|tar": Set the output type in `--output` option. "None" type allows to create files in steps of the docker. Default: "tar" (normally it should be converted to sqfs or ext2 to be used by the CM). Note that the none output type normally creates a dir at he the `dest=` location, so the flash drive itself should be the file inside the location.

for `builder = "raw"`.
- **builder** = "raw": create a raw empty drive that simply has length property. E.g. `--flash-drive=label:state,mount:false,user:dapp,length:67108864`. Useful for state flash drives (like honeypot)
- **length** = "size str": option used with `builder=raw` that should be converted to bytes

for `builder = "volume"`.
- **builder** = "volume": mount a local directory (requires `directory = "path"`) option as volume to customize the rootfs in shell mode.

### cli

It would be good to allow any cartesi.toml configuration to be customized/overiden when using the `cartesi build`/`cartesi shell` commands (e.g. `cartesi shell --machine-config "network=true" --machine-config "workdir=/mnt/app" --drive-config "app.mount=/mnt/app"`).

brunomenezes:
[...]

PR #498, opened 2026-06-30
Add env and env_file support to machine configuration

tuler:
## Motivation

Environment variables are typically used to differentiate between multiple build or test scenarios. One classic examples are testnet vs devnet vs mainnet, but there may be other scenarios.

Currently the build process only supports environment variables defined inside the Dockerfile used by the cartesi build process. But the same variables defined in the Dockerfile may be necessary outside the cartesi build process. For example in a host build process that is not related to the cartesi build process at all. So in this case there is a duplication of environment variables values, which hurts maintenance.

By having `.env` files support we can maintain a single source of truth for environment variables and use it in both the cartesi build workflow and other workflows.

## Summary
This PR adds support for injecting environment variables into the Cartesi Machine through two new configuration options in the `[machine]` section of `cartesi.toml`: an `env` table for inline key/value pairs and an `env_file` option pointing to a `.env` file.

## Key Changes
- **New configuration options**: Added `env` (Record<string, string>) and `envFile` (optional string) to `MachineConfig`
- **TOML parsing**: Implemented `parseStringRecord()` to parse TOML tables into environment variable records, with automatic coercion of non-string scalar values (numbers, booleans) to strings
- **`.env` file parsing**: Added `parseEnvFile()` function that parses `.env` files with support for:
  - Blank lines and `#` comments
  - Optional `export` prefix
  - Single/double quoted values
  - Whitespace trimming
  - Preservation of `=` signs in values
- **Environment variable precedence**: Updated `bootMachine()` to merge environment variables with proper precedence (lowest to highest):
  1. Docker image `ENV` (when `use_docker_env` is enabled)
  2. Variables from `env_file`
  3. Variables from the `env` table
- **Error handling**: Added `InvalidEnvError` exception class for invalid env configurations
- **Comprehensive tests**: Added unit tests covering all parsing scenarios and edge cases

## Implementation Details
- Environment variables are merged into a single map before being passed to `cartesi-machine`, allowing later sources to override earlier ones
- The `.env` file parser is robust and handles common `.env` file formats
- Non-string values in the `env` table are automatically converted to strings since environment variables are always strings
- The implementation maintains backward compatibility with existing configurations

https://claude.ai/code/session_016vnD51QAGwAvqaZuEZJypr

### DIFF
--- a/apps/cli/src/machine.ts
+++ b/apps/cli/src/machine.ts
@@ -1,3 +1,5 @@
+import dotenv from "dotenv";
+import fs from "node:fs";
 import type { Config, DriveConfig, ImageInfo } from "./config.js";
 import { cartesiMachine } from "./exec/index.js";
 import type { ExecaOptionsDockerFallback } from "./exec/util.js";
@@ -34,6 +36,8 @@ export const bootMachine = (
     const { machine } = config;
     const {
         assertRollingTemplate,
+        env: envConfig,
+        envFile,
         maxMCycle,
         ramLength,
         ramImage,
@@ -42,12 +46,29 @@ export const bootMachine = (
         user,
     } = machine;
 
-    // list of environment variables of docker image
-    const env = useDockerEnv ? (info?.env ?? []) : [];
-    const envs = env.map((variable) => {
-        const [key, value] = variable.split("=");
-        return `--env=${key}="${value}"`;
-    });
+    // environment variables injected into the cartesi-machine ENV, by
+    // increasing precedence: docker image ENV, env_file, then the env object
+    const envMap: Record<string, string> = {};
+
+    // environment variables of docker image
+    if (useDockerEnv) {
+        for (const variable of info?.env ?? []) {
+            const [key, value = ""] = variable.split("=");
+            envMap[key] = value;
+        }
+    }
+
+    // environment variables from an .env file
+    if (envFile) {
+        Object.assign(envMap, dotenv.parse(fs.readFileSync(envFile)));
+    }
+
+    // environment variables explicitly defined in the config
+    Object.assign(envMap, envConfig);
+
+    const envs = Object.entries(envMap).map(
+        ([key, value]) => `--env=${key}="${value}"`,
+    );
 
     // check if we need a rootfstype boot arg
     const root = config.drives.root;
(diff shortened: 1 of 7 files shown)


apps/cli/src/machine.ts
```diff
@@ -42,12 +46,33 @@ export const bootMachine = (
...
+
+    // environment variables of docker image
+    if (useDockerEnv) {
+        for (const variable of info?.env ?? []) {
+            const eq = variable.indexOf("=");
+            if (eq === -1) {
+                envMap[variable] = "";
+            } else {
+                envMap[variable.slice(0, eq)] = variable.slice(eq + 1);
+            }
```
brunomenezes: @tuler,
These slices are odd. Could it be done by using a destructuring based on the assumptions of the `variable` content?

e.g.
```js
const [key, value=""] = variable.split("=");
envMap[key] = value;
```

brunomenezes:
The PR description is strongly on the "how", but I don't understand clearly the "why".  Does it help solve an issue Lyno opened? 
* #477

tuler:
> The PR description is strongly on the "how", but I don't understand clearly the "why".

Added a motivation section to the PR description.

apps/cli/tests/unit/config/fixtures/full.toml
```diff
@@ -11,6 +11,11 @@
 # use_docker_env = true
 # use_docker_workdir = true
 # user = "dapp"
+# env_file = ".env" # optional. path to a .env file with environment variables
+
+# [machine.env] # optional. environment variables injected into the machine
+# FOO = "bar"
+# LOG_LEVEL = "debug"
```
brunomenezes, line 18: @tuler, 
I am curious and just trying to understand how to use it. Is the idea to plug in anything from the environment variable, like arbitrary values that only make sense to what is running inside, or is it to change the behaviour of the machine, like the `rollups_node` env vars?  If the latter would be good to have a list of behaviour-changing ones as they're specific and "reserved" in some sense.

PS: When I build dapps, the default configurations so far work for me, so I am trying to grasp the use. Even though you mentioned the **motivation** as the workflows and tests on different nets (dev, test ,main), it is not clear to me what these cases are; I mean what users are actually doing.

tuler: it's for any arbitrary env var the application may use to configure the application logic.
examples for a financial application:
COLLATERAL_TOKEN_ADDRESS
PRICE_FEED_ADDRESS

brunomenezes (approved):

PR #499, opened 2026-07-01
feat: print build logs on verbose mode

brunomenezes:
# Summary
Currently, the `cartesi build` command, when using Docker under the hood, is in silent mode even when the `--verbose` flag is passed. If the build takes a while to happen, the user has no idea what is going on unless they realize to go into `docker desktop -> builds -> click tab active build -> tab logs` and have a look at the current situation.  However, in a CI environment, that is not an option. 

Now, by passing the `--verbose` flag, you will be able to see the logs of the Docker build. 

**PS: We're dogfooding by enabling it on our own integration tests that require an application to be built before the cases are run against it.**

### DIFF
--- a/apps/cli/src/builder/docker.ts
+++ b/apps/cli/src/builder/docker.ts
@@ -1,13 +1,16 @@
 import { execa } from "execa";
 import fs from "fs-extra";
 import path from "node:path";
+import tmp from "tmp";
 import type { DockerDriveConfig } from "../config.js";
 import { genext2fs, mksquashfs } from "../exec/index.js";
+import type { Reporter } from "../exec/util.js";
+import type { BuildxMetadata } from "../types/docker.js";
 
 type ImageBuildOptions = Pick<
     DockerDriveConfig,
     "buildArgs" | "context" | "dockerfile" | "tags" | "target"
-> & { destination: string; dockerfileContent?: string };
+> & { destination: string; dockerfileContent?: string; reporter?: Reporter };
 
 type ImageInfo = {
     cmd: string[];
@@ -26,6 +29,7 @@ const buildImage = async (options: ImageBuildOptions): Promise<string> => {
         destination,
         dockerfile,
         dockerfileContent,
+        reporter,
         tags,
         target,
     } = options;
@@ -43,7 +47,7 @@ const buildImage = async (options: ImageBuildOptions): Promise<string> => {
         "--output",
         `type=tar,dest=${destination}`,
         "--progress",
-        "quiet",
+        reporter ? "plain" : "quiet",
     ];
 
     // set tags for the image built
@@ -52,16 +56,38 @@ const buildImage = async (options: ImageBuildOptions): Promise<string> => {
     // set build args
     args.push(...buildArgs.flatMap((arg) => ["--build-arg", arg]));
 
+    // use --metadata-file to capture the image ID from json format file, so stdout can be safely
+    // ignored regardless of what --progress mode outputs there
+    const tmpFile = tmp.tmpNameSync();
+    args.push("--metadata-file", tmpFile);
+
     if (target) {
         args.push("--target", target);
     }
 
     args.push(context);
 
-    const { stdout: imageId } = await execa("docker", args, {
-        input: dockerfileContent,
-    });
-    return imageId;
+    if (reporter)
+        reporter(`Building docker image with args: ${args.join(" ")}`);
+
+    const proc = execa("docker", args, { input: dockerfileContent });
+    if (reporter) {
+        proc.stderr?.on("data", (chunk: Buffer) => {
+            for (const line of chunk.toString().split("\n")) {
+                if (line.trim()) reporter(line.trimEnd());
+            }
+        });
+    }
+    await proc;
+    const metadata = JSON.parse(
+        fs.readFileSync(tmpFile, "utf-8"),
+    ) as BuildxMetadata;
+
+    return (
+        metadata["containerimage.config.digest"] ??
+        metadata["containerimage.digest"] ??
+        ""
+    );
 };
 
 /**
@@ -101,6 +127,7 @@ export const build = async (
     sdkImage: string,
     destination: string,
     debug: boolean,
+    reporter?: Reporter,
 ): Promise<ImageInfo | undefined> => {
     const { format } = drive;
 
@@ -116,11 +143,13 @@ export const build = async (
             ...drive,
             destination: path.join(destination, tar),
             dockerfileContent: `FROM ${drive.image}`,
+            reporter,
         });
     } else {
         image = await buildImage({
             ...drive,
             destination: path.join(destination, tar),
+            reporter,
         });
     }
     const imageInfo = await getImageInfo(image);
@@ -134,6 +163,7 @@ export const build = async (
                     output: filename,
                     cwd: destination,
                     image: sdkImage,
+                    reporter,
                 });
                 break;
             }
@@ -143,6 +173,7 @@ export const build = async (
                     output: filename,
                     cwd: destination,
                     image: sdkImage,
+                    reporter,
                 });
                 break;
             }
--- a/apps/cli/src/commands/build.ts
+++ b/apps/cli/src/commands/build.ts
@@ -31,9 +31,20 @@ const buildDriveTask = (
     task: async (ctx, task) => {
         const { config, debug, destination } = ctx;
         const sdk = config.sdk;
+        const reporter = (line: string) => {
+            task.output = line;
+        };
+
         switch (drive.builder) {
             case "directory": {
-                await buildDirectory(name, drive, sdk, destination, debug);
+                await buildDirectory(
+                    name,
+                    drive,
+                    sdk,
+                    destination,
+                    debug,
+                    reporter,
+                );
                 break;
             }
             case "docker": {
@@ -43,6 +54,7 @@ const buildDriveTask = (
                     sdk,
                     destination,
                     debug,
+                    reporter,
                 );
                 if (imageInfo && name === "root") {
                     // only set image info for root drive
@@ -55,7 +67,7 @@ const buildDriveTask = (
                 break;
             }
             case "tar": {
-                await buildTar(name, drive, sdk, destination);
+                await buildTar(name, drive, sdk, destination, reporter);
                 break;
             }
             case "none": {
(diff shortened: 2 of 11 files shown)


apps/cli/src/builder/docker.ts
```diff
@@ -52,16 +56,39 @@ const buildImage = async (options: ImageBuildOptions): Promise<string> => {
...
+        }
+        await proc;
+        const metadata = JSON.parse(
+            fs.readFileSync(tmpFile, "utf-8"),
+        ) as BuildxMetadata;
+
+        const imageName = metadata["image.name"]?.split(",")[0];
+        return imageName ?? metadata["containerimage.config.digest"] ?? "";
+    } finally {
+        fs.removeSync(tmpFile);
```
tuler: at `build.ts` we set `tmp.setGracefulCleanup()`, so we don't need to remove the file manually, hence no need for the try-finally block.

apps/cli/src/builder/docker.ts
```diff
@@ -52,16 +56,39 @@ const buildImage = async (options: ImageBuildOptions): Promise<string> => {
...
+                    if (line.trim()) reporter(line.trimEnd());
+                }
+            });
+        }
+        await proc;
+        const metadata = JSON.parse(
+            fs.readFileSync(tmpFile, "utf-8"),
+        ) as BuildxMetadata;
+
+        const imageName = metadata["image.name"]?.split(",")[0];
```
tuler: I'd stick with what is in the documentation, which is `"containerimage.digest"`
https://docs.docker.com/reference/cli/docker/buildx/build/#metadata-file

apps/cli/src/commands/build.ts
```diff
@@ -29,11 +30,23 @@ const buildDriveTask = (
 ): ListrTask<BuildContext> => ({
     title: `Building drive ${chalk.cyan(name)}`,
     task: async (ctx, task) => {
-        const { config, debug, destination } = ctx;
+        const { config, debug, destination, verbose } = ctx;
         const sdk = config.sdk;
+        const reporter = verbose
+            ? (line: string) => {
+                  task.output = line;
```
tuler: if the output is bound to the task, do we need to control verbosity at all, instead of always be like the verbose? Because it will hide once it's finished, no?

apps/cli/src/types/docker.ts
```diff
@@ -21,3 +21,26 @@ export type ServicePublisher = {
...
 
 export type PsResponse = ServiceStatus;
+
+/**
+ * A partial representation of the metadata produced by `docker buildx build --metadata-file`.
+ */
+export type BuildxMetadata = {
+    /**
+     * The container image digest (Image ID).
+     * Note: This may be undefined/null if using `type=local` or `type=tar`
```
tuler: we are controlling how buildx is called, which I believe is with
```
--output type=docker --output type=tar,dest=<file.tar>
```

apps/cli/src/builder/docker.ts
brunomenezes: I will remove `image.name` and leave the existing digest below. But the correct one is `containerimage.config.digest` here is the first build that was only using the one mentioned https://github.com/cartesi/cli/actions/runs/28516129205/job/84528221902#step:9:697. Looks like in the OCI image spec they're used for different purposes, and the `containerimage.config.digest` is the one we see when running the command `docker images` under the `ID` column.

apps/cli/src/commands/build.ts
brunomenezes: Not sure if I am following; what is your suggestion? Hide what exactly?

apps/cli/src/builder/docker.ts
brunomenezes: But that is a few calls above the stack, and the `build()` function is used elsewhere, so it may or may not set this "global" cleanup call. I think it makes sense to clean it in the same place where you create a resource.

apps/cli/src/builder/docker.ts
tuler: Hmm, that's weird. Will double check this later.

apps/cli/src/builder/docker.ts
tuler: But that is what `setGracefulCleanup` is for. Is there any other command that hits the `build` function?

apps/cli/src/commands/build.ts
tuler: I'm suggesting not having the`--verbose` in the build command at all, and always show the build log during the build, but not necessarily when it's over.

apps/cli/src/builder/docker.ts
brunomenezes: There are uses in tests based on the return from `find all references`.

apps/cli/src/builder/docker.ts
endersonmaia: `containerimage.digest` is the right one

`config` namespace if something else

apps/cli/src/builder/docker.ts
tuler: a call to `tmp.setGracefulCleanup()` in a `beforeAll` would do

apps/cli/src/commands/build.ts
brunomenezes: I did a fix-up. The `reporter` is always created, and I leave it to Listr2's renderers to decide how to display the outputs, i.e.  the`default` prints a single line (always updating and does the collapsing of messages at the end) and the `verbose` leaves the trail as expected.

apps/cli/src/builder/docker.ts
brunomenezes: @endersonmaia I linked the build above. Using that `containerimage.digest` causes problems as the Docker daemon reports that it is not able to find the image requested with the hash passed.  

For context, when using locally and running the tests, it runs fine, and upon inspecting the metadata, both `containerimage.digest` and `containerimage.config.digest` have the same SHA256,  but once pushed and the CI runs with QEMU, the story is different, and both options have different hashes, and the option discussed is the one that cause the error in the build link I shared above.

apps/cli/src/builder/docker.ts
brunomenezes: ok, that is done in a fix-up I force-push.

apps/cli/src/builder/docker.ts
endersonmaia: For me the test is also failing locally.

It works with `containerimage.digest`.

tuler: https://claude.ai/share/bd09c5ac-6a60-42d3-a4a9-90a7da857231

tuler: The image id is being used to run against `docker image inspect`, to get some data like command, entrypoint, env var, workdir, to be used during the cartesi machine build. I think that is the only reason, but it's worth to check.

So, based on the Claude discussion above it looks like `containerimage.config.digest` is the correct one.

brunomenezes: > For me the test is also failing locally.
> 
> It works with `containerimage.digest`.

If you can run that again locally after a `git pull --rebase`, I cleaned up the test fixup commits we did earlier. 

I included an order of precedence in the returns: `containerimage.config.digest` > `containerimage.digest`. 

### Why
What I understood so far is that locally, we use an embedded buildkit directly on the Docker daemon, so when it builds, it saves the image in the `docker image store`, but it behaves differently between macOS and Linux, as for me both properties are generated locally with the same hash, but on @endersonmaia it only generates the `containerimage.digest`.  **In the CI**, the `docker/setup-buildx-action` creates a new driver in a container, so inside the container there is a buildkit running standalone (i.e. buildkitd), with no `dockerd` inside it. Furthermore, it will generate the metadata file based on the OCI spec, and in that case the `containerimage.config.digest` is what we want.  Because of the `type=docker` argument, the built image was streamed to the host Docker as a tarball, unpacked and indexed in the host Docker images store. The `containerimage.digest` generated is essentially BuildKit's internal tracking ID of the build result in that `docker-container`, and the host Docker daemon never sees it nor stores it.


I could not find a configuration that would meet our needs, i.e., multiple outputs (type=docker and type=tar) and Cross-platform (riscv64) at the same time. For example, we can change the drive to be `docker` to run in the host or control BuildKit's version, but as far as I understand, neither will help in that situation and the `cartesi build` will fail in the **test** step.

cc: @tuler, @endersonmaia

tuler: What if we stop supporting these info coming from docker (CMD, ENTRYPOINT, ENV, WORKDIR), and use only what is defined in the `cartesi.toml`. Then maybe we don't even need the image id and the docker output.

brunomenezes: @tuler, I would like to take a closer look at that specific part later. Perhaps in a separate PR/branch.

tuler: But still I did not understand yet in what circumstances using `containerimage.digest` is correct.

brunomenezes: > But still I did not understand yet in what circumstances using containerimage.digest is correct.

When building with **the embedded buildkit via the local Docker daemon**, you should use `containerimage.digest`, as the Docker daemon only tracks single-platform images, and buildkit generates a "flat" local image format, so the build and manifest are treated as one and the same. 

Just to reiterate on the other side, when building with **the `docker-container`, buildkit operates independently of the Docker daemon's local "constraints** and, as I mentioned before, outputs a standard OCI multi-platform images structure. So in that case, we use the `containerimage.config.digest` as Buildkit maps the concrete `image ID` of the specific platform to this field.

tuler: So, based on some tests, here is my conclusion:

1) we should use `--iidfile` instead of `--metadata-file`. It's much simpler, we don't need to commit to any json structure, just read the image id from the produced text file. (this by itself don't solve the inconsistency problem yet, see point 2 below).

2) I think we should have more control over the buildx instance to use, to make the build process more consistent across platforms and docker installations. We can create a builder named `cartesi`, conditionally if it doesn't exist yet), and use that builder explicitly. This will give more control over the builder platform and the builder driver. It has been tested on macOS with Docker Desktop and on Linux with Docker Engine, with consistent results.

A builder can be created with the command below.
```
docker buildx create --driver docker-container --name cartesi --platform linux/riscv64
```

If it already exists and no `--append` is specified it exits with an error. Or it's also possible to run `docker buildx inspect --name cartesi`.

When building a builder can be used with `docker buildx build --builder cartesi`

tuler (approved):

PR #500, opened 2026-07-01
feat(sdk): add cartesi sequencer:v0.1.0-alpha.7

endersonmaia:
The sequencer binary is necessary in the SDK image if the developer want's to use the sequencer to send inputs instead of relying only on the InputBox.

cartesi/cli still needs to be adapted to use this service.


endersonmaia:
Converting to draft since there's a new version of the sequencer being prepared, so we don't need to bump it again.

PR #508, opened 2026-08-04
chore(cli):  Update cartesi-machine-stored-hash test case.

brunomenezes:
## Summary
This PR changes the test assertion for `cartesi-machine-stored-hash` cases to verify that a valid hash string is present, rather than asserting against a fixed hash value.

Because build environments can produce slight variations in the resulting output hash, strict hash comparison was causing test failures in CI/local environments—even when the dApp template code itself remained unchanged.

### DIFF
--- a/apps/cli/tests/integration/exec/cartesi-machine-stored-hash.test.ts
+++ b/apps/cli/tests/integration/exec/cartesi-machine-stored-hash.test.ts
@@ -2,6 +2,7 @@ import { afterAll, beforeAll, describe, expect, it } from "bun:test";
 import { writeFileSync } from "node:fs";
 import path from "node:path";
 import tmp from "tmp";
+import { isHash } from "viem";
 import { cartesiMachineStoredHash } from "../../../src/exec";
 import {
     createTemporaryCartesiApplication,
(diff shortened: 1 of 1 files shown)


endersonmaia:
> even when the dApp template code itself remained unchanged.

We should aim a reproducible build.

Isn't there any thing we can do to ensure this.

brunomenezes:
> > even when the dApp template code itself remained unchanged.
> 
> We should aim a reproducible build.
> 
> Isn't there any thing we can do to ensure this.

I am not sure, as from time to time we are changing the application-template Dockerfile, as it will eventually fail to build and updates are necessary. If I understood it right, the machine hash-generated factor, not just the code.

endersonmaia (approved):

tuler:
It’s nice to have a check of the hash. Maybe the application template repo should have in his repo the expected hash, and its CI do the check.

And the CLI just check if a hash is generated, and trust the application template repo is doing the stricter test.

tuler (approved):

brunomenezes:
> It’s nice to have a check of the hash. Maybe the application template repo should have in his repo the expected hash, and its CI do the check.
> 
> And the CLI just check if a hash is generated, and trust the application template repo is doing the stricter test.

I think that is the way due to these unrelated updates from time to time upstream that affect us here.

PR #509, opened 2026-08-05
feat(cli): replace the cartesi-machine spawn with @cartesi/machine

tuler:
Contributes to #72.

That issue asks for the external programs the CLI drives to stop being spawned binaries, and notes that _"instead of spawning binaries another possibility is to build NodeJS bindings to native code components"_. This PR takes that route for the emulator.

| Program | Before | After |
|---|---|---|
| `cartesi-machine` | `execa`, falling back to `docker run` in the SDK image | [`@cartesi/machine`](https://github.com/cartesi/rollups-ts) |
| `cartesi-machine-stored-hash` | `execa`, falling back to `docker run` in the SDK image | `@cartesi/machine` |

`xgenext2fs` gets the same treatment in **#517**, stacked on this branch. `mksquashfs`, `cartesi-rollups-cli` and `cartesi-rollups-node` are untouched by either, so the remaining work on #72 is the Node Unit's binaries.

> [!NOTE]
> The branch name still says `xgenext2fs-cartesi-machine` — it carried both changes before the split, and GitHub does not allow repointing a PR's head branch. Only the machine commit is here now.

## What changed

**`src/machine.ts`** — the substantial part. The CLI used to hand `cartesi-machine` a command line and let it assemble the machine configuration. It now does that translation itself, transcribed from `cartesi-machine.lua` v0.21.0:

- default `dtb.bootargs` from the emulator, with `machine.boot_args` appended
- `dtb.init`: the splash, then `dev=$(flashdrive <label>)` + mount + chown per non-root drive, then `export K="V"`, `WORKDIR`, `USER` — in the order the old argv produced
- flash drives with `root` first, so it lands on `/dev/pmem0` regardless of the order drives appear in `cartesi.toml`
- virtio console, `console=hvc1` and `iunrep=1` for `cartesi shell`

`buildMachineConfig` is a pure function so this is unit-testable without running a machine; `src/exec/cartesi-machine.ts` holds the run loop (automatic yields acknowledged, console I/O resumed, guest exit code read from `htif_tohost_data`).

**`src/images.ts`** (new) — with no SDK image to take `linux.bin` from, the default `ram_image` is fetched from the pinned `cartesi/machine-linux-image` v0.21.0 release on first use, verified against its SHA-256 and cached under `$XDG_CACHE_HOME/cartesi/images`. A `CARTESI_IMAGES_PATH` directory holding the image is used when set, and `machine.ram_image` still wins over both.

## Breaking

- **Machine hashes change.** `@cartesi/machine` links against emulator 0.21, the SDK image pins 0.20. Applications need redeploying.
- **Boot args are no longer double quoted.** The old code passed `--append-bootargs="<arg>"` with no shell in between, so the quotes ended up in the kernel command line. Building the configuration directly, reproducing that would have meant reproducing a bug.
- **The standalone binaries are gone.** A Bun single-file executable has no `node_modules`, and the addon resolves its platform `.node` at runtime — a host-native `--compile` fails too, so cross-compiling four targets was never going to work. The compile step in `build.ts` and the upload step in `release.yaml` are removed here. **The homebrew formula in `cartesi/homebrew-tap` needs to install from npm rather than the release tarball.** Happy to revert this bit and solve distribution differently if you'd rather keep the binaries.

## Testing

- 164 unit tests pass; `tests/unit/machine.test.ts` is new and covers the configuration translation (mount points, drive ordering, environment precedence, workdir/user, interactive mode, `parseMemorySize`).
- The version-reporting test for `cartesi-machine` moved from `tests/integration/` to `tests/unit/`, since it no longer needs Docker.
- Ran the built CLI end-to-end on a sample project: kernel downloaded and cached, machine created and stored, and the stored hash matched the final hash.
- `cartesi-machine-stored-hash.test.ts` asserted a literal hash that the emulator bump invalidates. It now asserts shape and stability — **worth pinning the real value once CI produces it.**

Not verified locally: a successful rollup application boot, which needs Docker with riscv64. The integration tests cover that path.


tuler:
### Note on the `@deroll` dependency

This PR depends on `@deroll/cm@alpha` and `@deroll/genext2fs@alpha`, which are personal-scope packages. That is expected to be temporary.

Per https://github.com/cartesi/rollups-ts/issues/133#issuecomment-5071569331:

> Next step is to bring `@deroll/cmio` here as `@cartesi/rollup` and `@deroll/cm` as `@cartesi/machine`.
> Will wait for machine-emulator 0.21.0 release for that.

So `@deroll/cm` → `@cartesi/machine` is already planned, and the release it was waiting on is the one this PR pins (`@deroll/cm@0.2.0-alpha.4` links against emulator 0.21.0). [Earlier in that thread](https://github.com/cartesi/rollups-ts/issues/133#issuecomment-5023267603) the per-platform packages are described as `@cartesi/machine-{darwin,linux}-{arm64,x64}`, _"currently published as `@deroll/cm`"_ — so the platform `optionalDependencies` get renamed alongside the main package.

Once those land, the migration here should be a rename rather than a rework: the API surface is the same, and the only imports to touch are

- `src/exec/cartesi-machine.ts`
- `src/exec/cartesi-machine-stored-hash.ts`
- `src/machine.ts` (types only)
- the `external` array in `apps/cli/build.ts`

`@deroll/genext2fs` is newer than that comment and isn't named in the plan, so it's worth deciding separately whether it also moves to `@cartesi/*` — it is the one dependency here under GPL-2.0-only (inherited from xgenext2fs), while the CLI is Apache-2.0.


---
_Generated by [Claude Code](https://claude.ai/code)_

apps/cli/tests/integration/exec/cartesi-machine-stored-hash.test.ts
```diff
@@ -29,29 +28,21 @@ afterAll(() => {
 });
 
 describe("cartesi-machine-stored-hash", () => {
-    it("should return the correct computed hash", async () => {
-        const machineHash = await cartesiMachineStoredHash.computeHash(".", {
-            forceDocker: true,
-            image: TEST_SDK,
-            cwd: machineDir,
-        });
+    it("should return the hash of a stored machine", async () => {
```
brunomenezes, line 31: I opened a PR regarding that
* #508

tuler:
### Migrated to `@cartesi/machine`

`@cartesi/machine@1.0.0-alpha.0` was published earlier today, so the swap my [previous comment](https://github.com/cartesi/cli/pull/509#issuecomment-5186788742) described as pending is done in 1abc6bb. `@deroll/cm` is gone from the dependency tree.

It was a rename, as expected — the two packages have an identical export surface (83 exports, same names; the only `.d.ts` difference is formatting), both link against emulator 0.21.0, and the platform packages carry the `@cartesi/machine-{darwin,linux}-{arm64,x64}` names from the issue thread. It also moves this dependency from `@deroll`'s Apache-2.0 to the official package under the same license.

Verified equivalence rather than assuming it: rebuilding the same machine (same drives, kernel, env, workdir, user) before and after produces the same root hash, `0x3231ba4754a106cf5c32bdec49d159823dced0e28c8e27565434c834fde94380`. Lint clean, 169 unit tests pass, and `build` + `hash` still work end-to-end against the bundled CLI.

`@deroll/genext2fs` is unchanged — still the open question from the previous comment.


---
_Generated by [Claude Code](https://claude.ai/code)_

